PR #46, opened 2015-08-24
changing the vernacular of 'errors' to 'results'

shellscape:
This is in preparation for the separation of warnings from errors, and the addition of errors as a separate result type. Passes `npm test`


jwilsson:
Awesome, thanks a lot!

I'll merge this in while we continue the discussion in #47.

Issue #44, opened 2015-08-20
Warnings vs Errors


shellscape:
Given the discussion around https://github.com/lesshint/lesshint/issues/43 errors being swallowed, I'd like to raise the question of warnings vs errors. jshint plops every result into an 'errors' object, but has a code on every result that differentiates error from warning. In jshint's case, it has `W` for warning, `I` for info, and `E` for error. I'm not suggesting lesshint follow suit, as I've never been incredibly keen on using single letters for error codes, but it would be a great addition to the project to differentiate.

Consider this lesshint gulp reporter: https://github.com/sagikazarmark/gulp-lesshint-stylish, which is transforming the results and marking everything back from lesshint as a warning.

Ideally everything that lesshint is returning as a result at this moment, sans parse error support, should be a warning. Once parse error support is added, those results should be errors. I'm not sure if `info` has any use here.

jwilsson:
I totally agree and have actually been thinking about this myself.

Another feature that could follow from this is the ability to specify the severity of a rule violation. Everything defaults to `warning` but if desired, mark some rule violations as `errors` and exit with a different code etc.

@kokarn thoughts, ideas?

shellscape:
:+1:  to the ability to override severity

kokarn:
I think @jwilsson is spot on with this and that we all agree that's something that could be useful.

I don't think we should follow jshint tho, add warnings/info to an error object seems a bit backwards and the one letter codes just seems a bit stupid. 
@shellscape Do you know why they do single letters?

I think that we should either return a single object with all warnings/error/info on it or a single object with three objects, each containing it's own type. 

The more i think about it i like the second alternative a bit more, would be easier to enumerate, could enable handling of specific types differently without having to loop over everything and so on.

shellscape:
@kokarn no clue why they used single letters - if I had to guess it was left over from jslint or just a style choice. I'm personally OK with an array of result objects, each with their 'state' property. Splitting that up into `{ warnings: [], errors: [] }` is a matter of stylistic preference at that point, or another way to map the data. 

I've started work on this on our fork and will submit PRs along the way.

jwilsson:
Closing this, since everything's fine and dandy in #47.

PR #47, opened 2015-08-24
Errors vs warnings

shellscape:
In reference to: https://github.com/lesshint/lesshint/issues/44

This is dependant upon https://github.com/lesshint/lesshint/pull/46

Defines warnings vs errors. Since linters aren't inheriting anything from a parent object and are duplicating a lot of code, the code to allow overriding of their result type will have to be pasted manually in each linter. This PR accounts for missing `type` property on each result and adds the default of `linter.resultType.warning` if it's not found.

This can be considered a minor semver change for the fact that it significantly adds to functionality, or a patch change since it contains no breaking changes.

### DIFF
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -3,8 +3,6 @@
 
     "excludedFiles": [],
 
-    "mapResults": false,
-
     "attributeQuotes": {
         "enabled": true,
         "style": "single"
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -3,6 +3,8 @@
 
     "excludedFiles": [],
 
+    "mapResults": false,
+
     "attributeQuotes": {
         "enabled": true,
         "style": "single"
--- a/lib/lesshint.js
+++ b/lib/lesshint.js
@@ -60,11 +60,29 @@ Lesshint.prototype.checkPath = function (checkPath) {
 
 Lesshint.prototype.checkString = function (input, checkPath) {
     var result;
+    var matches;
+    var lineNo;
 
     try {
         result = linter.lint(input, checkPath, this.config);
     } catch (e) {
-        result = [];
+
+        matches = e.message.match(/\#(\d+)$/);
+
+        // lineNo should never be 0, but in the event that gonzales-pe
+        // throws a different message, let's account for it. I couldn't
+        // find any instance that this would happen.
+        lineNo = matches.length > 1 ? parseInt(matches[1]) : 0;
+
+        result = [{
+            // errors from gonzales-pe only return line number at the moment,
+            // so we can safely assume column 1, as the entire line is affected.
+            column: 1,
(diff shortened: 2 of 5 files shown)


jwilsson:
Nice, thanks a lot for working on this!

Let's see if we can refactor the linters into inheriting something from a parent object (or something else to avoid a lot of the duplication.)

However, while I know @kokarn [likes a single object with arrays for each type](https://github.com/lesshint/lesshint/issues/44#issuecomment-133295632) I found the [way ESLint does it](http://eslint.org/docs/developer-guide/nodejs-api.html#linter) really nice. They expose plain objects in an array with a `fatal` property if a parse error is encountered.

With some inspiration from ESLint, our objects could look something like this:
```js
{
    column: 4,
    fatal: false,
    line: 1,
    message: 'Opening curly brace should be preceded by one space.',
    rule: 'spaceBeforeBrace',
    severity: 'warning',
    source: '.foo{'
}
```

What's the majority's opinion on this? If the majority's leaning towards something else we'll obviously go with that.

shellscape:
I don't mind `severity` over `type` at all. It's less ambiguous for sure. Do we need the `fatal` property?

jwilsson:
:+1: Yeah, I thought about the `fatal` property too, I think we can skip it.

Hmm, I guess there's some use for separate arrays in reporters. But I still think it's pretty easy to just filter out all `warning` objects from a large array. `Array.filter()` will handle that, and it will help us keep core a bit lighter.

shellscape:
OK. I'll edit to the new syntax and remove the mapping tomorrow.

shellscape:
I'm back on the task tonight. Should have an update shortly.

shellscape:
Note: This project needs specs around the file property in error results.

kokarn:
@jwilsson Is this ready to merge?

jwilsson:
@kokarn The code looks good, just a test as @shellscape mentioned. 

I'd also like to find a nice way to override the severity on each result. But perhaps we should look at that later? I have some other thoughts on the overall handling of linters too. I'll try to get a issue up for some discussion about all this in the coming days.

shellscape:
I'd say this can be merged then. The test I mentioned would be strictly for redundancy (to make sure no one removed a property) and can be part of another commit. I'll add it to our next sprint if it doesn't get done before then. I'd say it'd be prudent to restructure the linters before providing the overriding mechanism so you/I/we aren't duplicating effort.

jwilsson:
I'll see if I can get a test in during the weekend.

Anyhow, great work and thanks a lot for this @shellscape!

jwilsson:
@kokarn @shellscape Added a test in 35131a87efdfff78bc2ffd001c95abcff9361e7f.

shellscape:
:+1:

kokarn:
Nice

PR #48, opened 2015-08-24
Property Ordering by Alpha

srogers202:
Thanks for all the hard work guys!

I just ran into the question of ordering properties alphabetically. I noticed that the issue was brought up, so I put something together really quick. Let me know how it looks.

@jsdevel

### DIFF
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -66,6 +66,11 @@
         "style": "include_zero"
     },
 
+    "propertyOrdering": {
+        "enabled": true,
+        "style": "alpha"
+    },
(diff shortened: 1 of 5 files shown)


jsdevel:
+1

jwilsson:
Great job! Just some minor things as per the notes, then we're good to go.

srogers202:
@jwilsson Thanks for the feedback.

kokarn:
Really nice! Like it a lot :thumbsup:

shellscape:
Thanks for this one @srogers202 !

PR #54, opened 2015-08-29
fix(attributeQuotes): skip validation for value-less selector

CITguy:
Support value-less attributes (a.k.a. boolean attributes)


jwilsson:
Awesome! Thank you :tada:

Issue #56, opened 2015-08-31
Linter refactoring


jwilsson:
Continuing on our discussion in #47.

We have some ways in which we would like to change the linters to avoid code duplication.

### Disabling linters
Right now each linter checks whether or not it's disabled. In order to avoid lots of code duplication I'm proposing a move of this code to the main AST traversing, before the linter is even called. This might even result in the ability to just load the linters that are enabled in the user's `.lesshintrc`.

### Rule severity
We talked about how to override this in #47 while avoiding as much code duplication as possible. Unfortunately I haven't come up with any good ideas on how to specify the severity of a violation other than to include it in each linter. Perhaps someone else has? If so, tell us! Otherwise, I guess we'll have to include some sort of code in each linter.

kokarn:
I quite like the eslint way, to specify it in the first argument for the linter. 
1 is warning, 2 is error.

```json
{
     "space-in-parens": [
          1,
          "always",
          {
               "exceptions": [
                    "{}",
                    "()"
               ]
          }
     ],
     "space-before-function-paren": [
          2,
          "never"
     ],
     "space-unary-ops": 2,
     "no-spaced-func": 2,
     "no-multiple-empty-lines": 2,
     "no-multi-spaces": 2
}

shellscape:
I'm planning on adding time for this to our next sprint, FYI.

jwilsson:
:+1:

jwilsson:
@shellscape :+1:

shellscape:
Created #63 to continue discussion.

PR #63, opened 2015-09-10
initial commit of linter refactor concept, for discussion only

shellscape:
This pull request is intended to be for discussion only. If consensus is reached, changes will be implemented clean on a new branch. Referencing: #56 

NOTE: specs have not been updated to accommodate these changes.


jwilsson:
I think it looks great! Awesome work!

I definitely think we should continue iterating on this. @kokarn what do you think?

kokarn:
Oh, forgot to check this out. Yeah, looks good as always.

shellscape:
Been a bit of a nutty week. I'm off the grid until Monday, but I'll make an effort to get this wrapped up on Tuesday when I return.

shellscape:
Since discussion has concluded, working on the actual branch for the refactor.

Issue #62, opened 2015-09-08
Catch errors


kokarn:
We really need to look at how we catch errors. 

Right now the same output is produced if lesshint fails a file or if it succeeds with 0 errors which is really not good.

jwilsson:
@shellscape do you think it's enough to check `matches == null` at [4e01089fac#70](https://github.com/lesshint/lesshint/blob/ce3e9d60a5828323b50151668d3d9118d2414dfb/lib/lesshint.js#L70) and exit with `e.message` (or something) when that evaluates to `true` or do we need something more "robust"?

shellscape:
@jwilsson that's an old commit, before we changed `type` to `severity`.  did you mean https://github.com/lesshint/lesshint/blob/master/lib/lesshint.js#L70?

I think we should have a more robust pattern to match the parse error message we're targeting from gozales-pe. It should be of local-to-gonzales-pe-type `ParseError` which we could check with a `typeof`. From there, if the error thrown from g-pe isn't a ParseError, then we rethrow. That's when reporters like in gulp's pipeline would handle the error thrown.

In summary: we should swallow and handle `ParseError`s from gonzales-pe, but rethrow other errors with a reasonable expectation that lesshint will exit nonzero at that point.

jwilsson:
@shellscape Yeah, I just wanted a persistent link and that was the first commit I found :)

Sounds like a good idea. I took a quick look at the full `Error` object thrown by `gonzales-pe`, a bit cleaned up:

```js
{
    line: 2,
    syntax: 'less',
    css_: 'a {\n    color red\n}\n',
    name: 'Parsing error',
    version: '3.0.0-31',
    message: 'Please check validity of the block starting from line #2',
    context: '
    1 | a {
    2*|     color red
    3 | }'
}
```

I guess we could check for one/some of these properties and use that to decide what to do with the error.

jwilsson:
I'll take a stab at this tonight. Will submit a PR so we're all in agreement.

kokarn:
Great! Ping me on IRC when done and i'll check it out tonight.

kokarn:
Added with https://github.com/lesshint/lesshint/commit/0f7b4dc9dc1e721a9c6cc088ee16784761f5a795

PR #68, opened 2015-09-14
Better handling and catching of errors.

jwilsson:
As discussed in #62. Lets make sure we don't try to parse all kinds of exception messages, just the ones thrown by `gonzales-pe`. All other messages should just be rethrown and handled further up, for example in the `CLI` or by `gulp-lesshint`.


kokarn:
This looks really good. I agree with the decisions about error handling as well.

As long as you are confident in the tests covering the most use cases and as soon as the coverage data comes back I think we should merge.

PR #69, opened 2015-09-20
Add decimalZero linter

### LINKED ISSUE #26: Available options/default value for trailingZero
Do we need any more options? Are we satisfied with the current default value? Any other thoughts?

Current options:
1. exclude_zero (default)
2. include_zero

Please cast a vote for one of the options or post your other thoughts or suggestions for options.

CITguy:
* added default config
* remove leadingZero linter
* remove trailingZero linter
* updated linters README

Closes #26


kokarn:
Looks really good dude!

Only thing i can think of is if this should be ```both``` as a default and then minifiers can remove the extra zeroes?

Thoughts?

CITguy:
I set the default to `leading` because it is what translates from the other two linters: `leadingZero: include_zero` and `trailingZero: exclude_zero`.  I think this is a sane default, given that it seems more natural to write `0.5em` than `0.50em`.

CITguy:
Additionally, would `0.250em` seem like a sane value alongside `0.50em`?

kokarn:
I think you are quite right. Let's merge this and keep rolling :)

Great work!

Issue #36, opened 2015-06-10
Available options/default value for propertyUnits


jwilsson:
Do we need any more options? Are we satisfied with the current default value? Any other thoughts?

Current options:
global: [] Array of allowed units (currently all CSS units)
properties: {} Object with unit overrides for specific properties

Please cast a vote for one of the options or post your other thoughts or suggestions for options.

kokarn:
I'd like a global exclude option actually, maybe a exclude list for properties as well.

The usecase for that is we only currently dislike the vh unit because of it's crappy behaviour on mobile, everything else is good.

That way it's forwards compatible as well.

CITguy:
I like the idea of having two ways to configure the units (whitelisting vs blacklisting).
Maybe change the config to be something like this...

```javascript
/* BLACKLISTING */
"propertyUnits": {
  "permit": "deny",
  "units": ["rem", "vh"], // these units will not be allowed anywhere
  // property-specific units
  "properties": {
    "min-height": ["vh"] // vh will be allowed in addition to globally allowed units
  }
}

/* WHITELISTING */
"propertyUnits": {
  "permit": "allow",
  "units": ["px", "%"], // only these units will be allowed
  // property-specific units
  "properties": {
    "min-height": ["vh"] // vh will be allowed in addition to globally allowed units
  }
}
```

The majority of the configuration will take place with the `permit` and `units` values. The `properties` option will only be used to handle the edge cases.

kokarn:
That's great and gives a lot of flexibility. :+1:

jwilsson:
@CITguy I really like the idea but I feel this has been sitting for a while and I don't think I will have the time in the near future to implement it. Do you have the time?

shellscape:
I think the `permit` based approach is too superfluous. Here's my proposal:

```
"propertyUnits": {
  "valid": ["px", "%"], // only these units will be allowed
  "invalid": ["em"], // not allowed under any circumstance

  "properties": {
    "min-height": ["vh"] // vh will be allowed in addition to 'valid' units
  }
}
```

kokarn:
This is now solved thanks to @shellscape

PR #77, opened 2015-10-07
propertyUnits: valid, invalid specification. addresses #36

shellscape:
propertyUnits: valid, invalid specification. addresses #36

### DIFF
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -73,7 +73,7 @@
 
     "propertyUnits": {
         "enabled": true,
-        "global": [
+        "valid": [
             "ch", "em", "ex", "rem",                 // Font-relative lengths
             "cm", "in", "mm", "pc", "pt", "px", "q", // Absolute lengths
             "vh", "vw", "vmin", "vmax",              // Viewport-percentage lengths
--- a/lib/linters/README.md
+++ b/lib/linters/README.md
@@ -334,16 +334,18 @@ and the `properties` option can be used to fine tune units for each property.
 
 *Note: Shorthands are not supported by the `properties` option. For example, to specify units for `margin`, all margin-* properties must be specified.*
 
-Option       | Description
------------- | ----------
-`global`     | Allowed units (by default all units are allowed)
-`properties` | Object with property names and allowed units (empty by default)
+Option        | Description
+------------- | ----------
+`valid`       | Allowed units (by default all units are allowed)
+`invalid`     | Units not allowed under any circumstances. Overrides all other options.
+`properties`  | Object with property names and allowed units (empty by default)
 
 ```js
 "propertyUnits": {
-    "global": ["rem", "vw"], // These units are allowed for all properties
+    "valid": ["rem", "vw"], // These units are allowed for all properties
+    "invalid": ["pt"],      // The 'pt' unit is not allowed under any circumstances
     "properties": {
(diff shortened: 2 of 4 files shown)


kokarn:
This looks good and because you are so quick with the PRs i'd like to merge.
Mind having a look on the failing tests?

shellscape:
@kokarn no idea what's up with that. they were passing locally.

shellscape:
updating to gonzales-pe @ 3.0.0-beta was the cause of some of the errors. I reverted the commit.

kokarn:
Can you update the PR?

shellscape:
@kokarn in what way? specs are passing locally. I don't know why either services are failing.

kokarn:
Hmm, I guess we need to specify the correct version of gonzales-pe. Wouldn't the tests failing on Travis and Appveyor mean that it's using the wrong version?

shellscape:
I reverted the commit that was using 3.0.0-beta, and it's back to 3.0.0-31, which is what master is using.

kokarn:
Hmm, this is weird. There is only the one commit for this PR so it shouldn't fail :/

shellscape:
For whatever reason, I had to blow away and reclone to start seeing the failing specs. Anyhow, it's due to the propertyUnits change. There's one more error being thrown on the checkDirectory specs:

```
{ column: 19,
      line: 2,
      linter: 'propertyUnits',
      message: 'Unit "px" is not allowed for "margin-bottom".',
      file: 'vendor.less',
      source: '    margin-bottom:10px;',
      severity: 'warning' }
```

I ran into this yesterday with our internal linting. Because propertyUnits is on by default, and by default there are no units specified, **all** units are invalid.

I'm thinking this is a candidate for a disabled-by-default linter. Thoughts?

kokarn:
I'd say the defaults are messed up, not that it should be disabled.

A little googling gave me [this list](http://www.w3.org/TR/css3-values/)
`em`, `ex`, `ch`, `rem`
`vw`, `vh`, `vmin`, `vmax`
`cm`, `mm`, `q`, `in`, `pt`, `pc`, `px`
`deg`, `grad`, `rad`, `turn`
`s`, `ms`
`Hz`, `kHz`
`dpi`, `dpcm`, `dppx`

My suggestion is to enable `deg`, `rad`, `s`, `dpi`, `dppx`, `em`, `rem`, `vw`, `vh` and `px` and disable the rest.

shellscape:
Do any of our other linters follow that pattern? (Of having a linter option populated with defaults)

shellscape:
Ah ignore that. Just found this in defaults.json:

```
    "propertyUnits": {
        "enabled": true,
        "global": [
            "ch", "em", "ex", "rem",                 // Font-relative lengths
            "cm", "in", "mm", "pc", "pt", "px", "q", // Absolute lengths
            "vh", "vw", "vmin", "vmax",              // Viewport-percentage lengths
            "deg", "grad", "rad", "turn",            // Angle
            "ms", "s",                               // Duration
            "Hz", "kHz",                             // Frequency
            "dpi", "dpcm", "dppx",                   // Resolution
            "%"                                      // Other
        ],
        "properties": {}
    },
```

So I missed the defaults config file. Fixing that...

kokarn:
Looks like a missed `%` as well :)

kokarn:
Jay! Great work @shellscape.

PR #84, opened 2015-10-20
Skip url @import statements

### LINKED ISSUE #80: importPath doesn't handle URIs without quotes
First reported by @alferov over at lesshint/gulp-lesshint#1.

It looks like we're a bit to naïve when checking URIs in `importPath`. We always assume a string which of course isn't always the case.

This is a great candidate for a first issue, on line 24 of https://github.com/lesshint/lesshint/blob/master/lib/linters/import_path.js we need to check for both strings and "raw" URIs and make sure whichever one we find is the value being checked.

rmbl:
This MR skips all @import statements with url()s. Should fix #80.

### DIFF
--- a/lib/linters/import_path.js
+++ b/lib/linters/import_path.js
@@ -21,6 +21,11 @@ module.exports = {
             return null;
         }
 
+        // Check if string value exists and exit if its an url instead
+        if (!node.contains('string') && node.contains('uri')) {
+            return null;
+        }
+
         value = node.first('string');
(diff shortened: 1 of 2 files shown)


jwilsson:
Great! I'm gonna merge this but I still think we need to check for `@import`s without quotes.

rmbl:
Yeah, that would be better than just skipping it. I might give it a try later.

kokarn:
No stress! Any contribution is very welcome!

PR #85, opened 2015-10-20
Skip variable declarations in propertyUnits linter

rmbl:
This MR skips variable declarations inside of classes similar to the example below.

```less
ul {
    @var-name: 25px;
}
```


jwilsson:
Awesome, thanks man!

PR #86, opened 2015-10-20
Lint quotes in @import urls

rmbl:


jwilsson:
Hmm, I think the real issue here is the fact that we're only checking strings and not `url()`s without quotes in the `importPath` linter. We shouldn't need to change anything else.

`importPath` should do the following:
```less
@import "foo.less"; // Check this
@import url(foo.less); // Check this
@import url("foo.less"); // Check this
@import url("http://example.com"); // Ignore this
@import url(http://example.com); // Ignore this
```

I still agree with #84, that absolute URLs should be ignored, since Less will leave these alone when compiling anyway (and one can't control file names etc. of external resources anyway).

rmbl:
This MR is only for urlQuotes linter. And I think its valid to lint all possible occurences of url() statements in this linter.

Your concerns regarding the importPath linter are correct I think. But skipping absolute URLs would be part of a different MR and wouldn't validate the quotes.

jwilsson:
I'm sorry, I misunderstood you. My bad.

I got confused since you referenced this PR in #80 since the issue there is another (`@import` statements without quotes aren't checked at all, perhaps I was unclear). Anyway, you're absolutely correct. `urlQuotes` should definitely check URIs in `@import` statements.

Thanks for the fix!

rmbl:
Then I probably misunderstood the ticket :)

PR #87, opened 2015-10-21
Ensure that mixins are on seperate lines

rmbl:
I extended the singleLinePerProperty linter to also include the 'include' and 'extend' types of gonzales-pe as those seem to be matched by ```.mixin();```. If you would like to include more or less types, please let me know.

### DIFF
--- a/lib/linters/single_line_per_property.js
+++ b/lib/linters/single_line_per_property.js
@@ -9,18 +9,20 @@ module.exports = {
         var results = [];
         var self = this;
 
-        node.forEach('declaration', function (element, index) {
-            var previousValid = self.isPreviousValid(node, element, index);
-            var nextValid = self.isNextValid(node, element, index);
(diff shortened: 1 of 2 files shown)


jwilsson:
I actually thought about this earlier today and I think we should add `variable` to the list.

I'm also thinking whether we should add `selector` and `atrule` too or if that's overkill. @kokarn what do you think and can you come up with anything else?

kokarn:
This is everything that should be on a separate line, correct?

Reading through the docs, the ones i find are 
functions
```less
.function {
    //some stuf
}
```
Detatched rulesets
```less
// declare detached ruleset
@detached-ruleset: { background: red; };

// use detached ruleset
.top {
    @detached-ruleset(); 
}
```
and loops. 
But those should probably all be covered by this already, right?

rmbl:
Okay, I updated the merge request and added tests for variables and detached rulesets. Variables worked already but I had to check ```atrules``` for the detached ruleset.

Issue #88, opened 2015-10-22
Add test to make sure we don't miss any console.logs in tests


kokarn:

jwilsson:
I'm not sure I follow, the `console.log`s in 0756cba7f1ee68ce0a05c4d89862bf8ed474416d were just stray ones which I forgot to remove.

Are you thinking of some sort of JSHint/JSCS check for `console.log`s in tests?

kokarn:
Yeah, that should work. 
As long as it stops them from showing up in the commited code where they shouldn't be :)

CITguy:
For a project I work on, we use a script to be run on the git pre-commit hook, so that it won't allow you to commit if there are debugging statements still present in the code. The hook installs itself via an NPM postinstall. Here's a gist that might be useful: https://gist.github.com/CITguy/1138a1701b9ffbf4e091

jwilsson:
@CITguy looks pretty nice, thanks for sharing!

I'll see if I can work something out.

jwilsson:
We talked about using Gulp and running all our tests and linting through it, and I figured we can check for stray `console.log()`s there as well (that way it'll be automatically run together with `npm test`). 

We just need to make sure it doesn't abort the tests while developing something. I'll work something out tonight.

shellscape:
:+1:

PR #92, opened 2015-10-31
Use Gulp for testing

jwilsson:
This is a first rough Gulp setup (and a test to see if it works on Travis). 

I just wanted to get this out there to iterate on, I'm also going to add a task to find `console.log()` calls as per #88.

Update: I just noticed that I based this off the wrong branch, I'll see if i can fix it when we merge it.


shellscape:
This looks great. My only suggestion would be to use:

```
expect(result).to.be.undefined;
```

instead of:

```
expect(result).to.equal(undefined);
```

jwilsson:
I wanted to do it like that but JSHint didn't like it (I guess we can turn it off since it's allowed in ES5). I'll look at it tonight.

shellscape:
At the very least we can disable just for those files as an inline comment in the header.

PR #98, opened 2015-11-17
Replace gonzales-pe with updated fork

jwilsson:
I'm a bit unsure whether we should pull it straight from Git or include the whole library in our source.

Pros:
- Bug fixes can be published straight away.
- Including the whole lib in our source will probably be quite large.
- ???

Cons:
- We might accidentally break things with a bad push.
- ???


shellscape:
We use gulp-lesshint, lesshint, and gonzales-pe from a git ref to our forks without any issues. I'd recommend that as the best way to go.

Issue #50, opened 2015-08-28
Add validateIndentation


CITguy:
My team uses 4 spaces per indentation. It would be nice to have a linter to check that indentation is correct. Something along the lines of the following (with suggested defaults):

#### syntax
```
"validateIndentation": {
  "enabled": true,
  "style": "spaces",
  "quantity": 4
}
```

#### options
**style** (string)

* spaces (default)
* tabs

**quantity** (integer) number of spaces/tabs per indentation

* 2 (default)
* user defined value

jwilsson:
I think this will be quite easy once [e8bdc3be2e34ff369f3b0f8ee699a1262a97b8a0 in `gonzales-pe`](https://github.com/tonyganch/gonzales-pe/commit/e8bdc3be2e34ff369f3b0f8ee699a1262a97b8a0) gets published on npm. 

We should be able to multiply the `level` argument with each `column` property in the AST. And then check that value against the `quantity` value multiplied with `level`.

kokarn:
I recently got a file that looks something like this
```
.class {
    property: value;

    next-property: value;

    .otherClass {
          property: value;
    }

    third-property: value;
}
```

Maybe newlines and all properties on top should be in this linter as well? Or separate?

CITguy:
Well, this is an interesting case.  What should the CSS look like?

## What should the LESS look like?

### Apply Linting to Properties

#### Option 1.0 - No blank lines before or after property definitions
```less
.class {
  property: value;
  next-property: value;
  .otherClass {
    property: value;
  }
  third-property: value;
}
```

#### Option 1.1 - Blank line before property definition
```less
.class {

  property: value;

  next-property: value;
  .otherClass {

    property: value;
  }

  third-property: value;
}
```

#### Option 1.2 - Blank line after property definition
```less
.class {
  property: value;

  next-property: value;

  .otherClass {
    property: value;

  }
  third-property: value;

}
```

#### Option 1.2 - Blank lines around property definition
```less
.class {

  property: value;

  next-property: value;

  .otherClass {

    property: value;

  }

  third-property: value;

}
```

### Apply Linting to Style Block

### Option 2.0 - Blank line before block
```less
.class {
  property: value;
  next-property: value;

  .otherClass {
    property: value;
  }
  third-property: value;
}
```

### Option 2.1 - Blank line after block
```less
.class {
  property: value;
  next-property: value;
  .otherClass {
    property: value;
  }

  third-property: value;
}
```

[...]

PR #107, opened 2015-12-06
First iteration of validateIndentation linter

jwilsson:
A first iteration of a `validateIndentation` linter.

Known issues:
* The closing bracket of a block isn't checked.
* Nested blocks will be reported even if the indentation is correct (can be seen when running from the CLI for example).

Questions/feedback wanted:
* Default value, we talked about `2` as the default indentation level in #50. Should we go with this or is `4` a better default?
* Name for the above "level" option. What should we call it? Suggestions:
    * `quantity` (suggested in #50)
    * `size`
    * `length`
    * `width`

Tests are failing due to incorrect indentation, will be fixed when we're in agreement. If anyone wants to take a stab at the known issues, please do!


idelpivnitskiy:
Looks like it very useful linter!
@jwilsson Any updates here?

jwilsson:
@idelpivnitskiy I'm afraid not, it turned out to be more work than I originally thought.

And it doesn't look like I'll have any time in the foreseeable future, hopefully it'll be less work once we replace our parser (tracking in #151).

If you wanna take a stab at it, you're more than welcome of course!

jwilsson:
Closing this since nothing will happen before we're done with the PostCSS conversion and this is too far behind `master` anyway.

Issue #105, opened 2015-12-05
[feature request] validate selector naming convention


AoDev:
Hello, 

I'd like to be able to validate selector naming.

Until now I have always been with people following the convention to use dashes / lowercase names but a coworker has been using camel case in his previous workplace. 

`btn-primary` vs `btnPrimary` 

Maybe a simple `lowercase` test is enough ?

At first I could see settings for:
* underscore (true | false)
* dash (true | false)
* case ('lowercase' | 'mixed')

kokarn:
This sounds insanely useful actually.

+1 from me

jwilsson:
I agree, sounds awesome.

@AoDev interested in a PR?

shellscape:
various common cases:

* sausage
* snake
* camel
* lower
* upper

AoDev:
@jwilsson I'll check the code to see how "hard" it is to prepare a PR. I am fighting multiple fronts and another just slows everything :P

AoDev:
@jwilsson I have started with it.

PR #108, opened 2015-12-08
feat(linter): selector naming convention #105

AoDev:
Allows to check for case, dashes and underscores.
Can be combined to approximate camelcase, train, snake case.
One can also add a white list with the excludes option.

Actual naming conventions like camel case or bem notation might be
a second step.

Issue  #105

### DIFF
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -93,6 +93,12 @@
         "allowWithAttribute": false
     },
 
+    "selectorNaming": {
+        "enabled": false,
+        "disallowUppercase": true,
+        "disallowUnderscore": true
+    },
+
     "singleLinePerProperty": {
         "enabled": true
     },
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -22,6 +22,7 @@ var linters = [
     require('./linters/property_ordering'),
     require('./linters/property_units'),
     require('./linters/qualifying_element'),
+    require('./linters/selector_naming'),
     require('./linters/single_line_per_property'),
     require('./linters/single_line_per_selector'),
     require('./linters/space_after_property_colon'),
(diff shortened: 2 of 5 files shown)


AoDev:
@jwilsson 

My approach towards this problem is to check for uppercase, dashes and underscores. These can be combined to have a more or less camel case, snake case, train case test.

I haven't found (yet) actual regexes to help with this. Especially with css selector hacks and browser prefixes I did not even want to try every edge case. I think that this approach is not perfect but will help in most cases. Let me know

jwilsson:
Hi,
Thanks for working on this!

I think it looks good, just two minor things:
* Could you rename `excludes` to simply `exclude`? For consistency with other linters.
* I think you missed to add a entry to `lib/config/defaults.json` with the default options.

Lets see if anyone else got any comments too, but I'm good.

shellscape:
Since this is one of the more complex linters in terms of options, I'd like to see the readme updated with combinations to accomplish what you were saying is possible. It's pretty straightforward to me, but might not be for others.

AoDev:
@jwilsson @shellscape I have pushed changes following your comments.

kokarn:
We use a technique called [Slightly obtrusive JavaScript](http://ozmm.org/posts/slightly_obtrusive_javascript.html) in some of our projects. I assume others use similar things, is that something that should be checked by this? 
Like detecting `.js-` prefix and invalidating.
I'm not sure, I think it might be better off in it's own linter but it's worth thinking about at least.

AoDev:
@kokarn

There are actual dedicated validators like this for example.
https://github.com/postcss/postcss-bem-linter

This particular namingConvention rule could be extended with more options. Including ready made validators like this bem linter.

But personally I see that as a second step that someone might want to take if interested.

kokarn:
I agree, bem linter should be a different linter.

:+1: for merge from me

jwilsson:
Thanks again @AoDev!

AoDev:
@jwilsson I just tried to use it with a real file and I think it needs a small fix.
  
```
var name = element.first('ident') && element.first('ident').content;

// it seems sometimes, name can be null. 
if (!name) {
    return false;
}
```

To be honest I checked how it was done in the id_selector rule and copied the same line to get the name.

Checking for the name like above is enough or is there something else I have missed?

PR #109, opened 2015-12-10
fix(selectorNaming): check if selector name is null

AoDev:


AoDev:
I realised that sometimes the selector name can be null if it's using an attribute selector like

`[type="text"]`

jwilsson:
Nice catch! Could you just add a test for it so we don't accidentally break it later.

AoDev:
@jwilsson I added the test.

jwilsson:
This is merged in master @ 55917c6b67e19debfb1cdcba8c90bd4d77b5b3c2. I removed the `ast.first()` stub, seemed a bit dangerous :smile: 

Will be published on npm in just a couple of minutes. Thanks for the patch! :tada:

AoDev:
@jwilsson Not sure why you think it's dangerous. Sinon is made for that and you just lost the test / coverage of this particular case. But thank you ; )

jwilsson:
@AoDev Sorry, I've should explained my reasoning better and I guess dangerous was the wrong word :) I didn't want to be dependent on the return value of a certain external method, I really just wanted to see whether a certain type of input would yield a `null` result.

I didn't notice any coverage changes for that case, but by all means, reopen this if you disagree :)

AoDev:
@jwilsson What I understand, is that you don't think that testing for null, is not enough. If you think that checking for `name === null` is not right then the problem was not the test but the linter.
Right now, it's like you don't trust the linter and you removed the test hoping that it will be fine.

I don't know all cases of output that can be generated in the AST. I only saw that when there is no name in a selector like for a attribute selector `[type="text"]`, it returns `null`, hence the check for a null value and the test that goes with it.

PR #115, opened 2016-01-30
Honor excludedFiles in the config file

cageygee3:
The current code tramples the excludedFiles setting in the config file
with either the value supplied on the command line or an empty array.
This fixes that; additionally, if excludedFiles is specified in both
the config file and on the command line, the union of both is used.

Tests have been updated/added to cover the different scenarios.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -70,8 +70,11 @@ module.exports = function (program) {
     try {
         config = configLoader(program.config);
         config = config || {};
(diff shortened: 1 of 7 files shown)


kokarn:
LGTM @jwilsson ? @shellscape ?

jwilsson:
Yup, looks great! 

Thanks a lot! :smile:

PR #135, opened 2016-02-18
docs(OPTIONS): remove dead links

CITguy:
`leadingZero` and `trailingZero` no longer exist. Removed their TOC links.


jwilsson:
Thanks!

PR #136, opened 2016-02-19
New Rule - Depth Level.

jcalvor:
New Rule - Depth Level. Sets the depth of nested styles in every parent
style.

### DIFF
--- a/lib/linters/depth_level.js
+++ b/lib/linters/depth_level.js
@@ -1,0 +1,73 @@
+'use strict';
+/*
+depthLevel
+
+Sets the depth of nested styles in every parent style.
+
+Option Description
+depth   1-n, 3(default)
+
+Invalid (depth 3)
+.foo {
+    margin-right: 0;
+    .foo-2 {
+      color: red;
+      .foo-3{
+        width: 100%;
+        .foo-4{
+          height: 100%;
+        }
+      }
+    }
+}
+
+Valid
+
+.foo {
+    margin-right: 0;
+    .foo-2 {
+      color: red;
+      .foo-3{
+        width: 100%;
+      }
+    }
+}
+*/
+var util = require('util');
+/*Validates that a less style wont containt more than 'n' nested styles, identifies the style parent that contains the error
+* Uses the Configuracion Param "Depth" for the number of permited levels that a style can be nested inside another style
+*/
+module.exports = {
+    name: 'depthLevel',
+    nodeTypes: ['block'],
+    message: "There shouldn't be more than '%s' levels deep from the style's parent, check the children's depth.",
+
+    lint: function deepLevelLinter (config, node) {
+      let depth = config.depth || 2;
+      let error = false;
+      node.eachFor('ruleset',function(childNode,i) {
+        let levelsTotal = 1;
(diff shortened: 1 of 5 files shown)


lib/linters/depth_level.js
```diff
@@ -0,0 +1,73 @@
...
+      if (error) {
+          return [{
+              message: util.format(this.message, depth)
+          }];
+      }
+    }
+};
+
+/*Checks how many nested classes have every node*/
+var countLevels = function (childNode,i, total) {
```
shellscape: argument spacing

lib/linters/depth_level.js
```diff
@@ -0,0 +1,73 @@
...
+  total += 1;
+  var block = childNode.first('block') || {};
+  block.eachFor('ruleset',function(subchildNode,i) {
```
shellscape: argument spacing

lib/linters/depth_level.js
```diff
@@ -0,0 +1,73 @@
...
+*/
+module.exports = {
+    name: 'depthLevel',
+    nodeTypes: ['block'],
+    message: "There shouldn't be more than '%s' levels deep from the style's parent, check the children's depth.",
+
+    lint: function deepLevelLinter (config, node) {
+      let depth = config.depth || 2;
+      let error = false;
+      node.eachFor('ruleset',function(childNode,i) {
```
shellscape: argument spacing

shellscape:
Looks good to me, but missing some things before the PR can be approved.

Need to create tests for the linter.
Need to create the readme for the linter and update the linters TOC to point to it as well.

You've got a bunch of linting errors in the code though. see https://travis-ci.org/lesshint/lesshint/jobs/110491786. Please run linting and tests locally as well.

jwilsson:
@shellscape beat me to it :)

It looks good, just the things he mentioned.

jcalvor:
O,k thanks for the feedback. I made some changes to fix the lint errors. I also created a test and updated the readme doc. Let me know if every is ok or if I need to make a change.

kokarn:
Only thing I can think of is that maybe this should be disabled as a default instead of 3? 
Thoughts?

CITguy:
Agree with @kokarn on disabled by default. However, I'd name it `validateIndentation` as suggested in #50.

CITguy:
Also, how is this different from #107?

kokarn:
Wow, yeah that's true. Is #107 ready to merge @jwilsson or should we go ahead with this?

jwilsson:
@CITguy @kokarn This PR checks the number of nested selectors while #107 (which isn't ready yet, it's a bit trickier then I originally thought) only checks your indentation.

It seems like the majority thinks this should be disabled by default, and I can see the reasoning behind that too, I think we should merge it and just set it to disabled.

kokarn:
If that's what you want and don't feel #107 is done, let's do that. Then maybe we can just improve on this at a later date.

jwilsson:
Huge thanks for this @jcalvor! :tada:

jcalvor:
Thank you guys! This is a great project and i hope this can help others.

Issue #55, opened 2015-08-30
Add ability to disable linting for a single line


CITguy:
Other linters/hinters have the ability to ignore linting on a single line via a trailing comment. It would be very useful to have this ability. An example...

My team discourages the use of `!important` but there are edge cases where it needs to be used.  As such, it would be useful to enable `importantRule` globally, but apply something like `// lesshint ignore:line` that would ignore linting for a specific line, so that I can allow `!important` in edge cases (bonus points if we can just ignore the single rule via `// lesshint ignore:importantRule`).

CITguy:
It should be trivial to implement `// lesshint ignore:line` as it would be a matter of removing lines from the incoming css *before* parsing with gonzales (similar to how we already [normalize line endings](https://github.com/lesshint/lesshint/blob/master/lib/linter.js#L76)).  This way, the entire line is completely excluded from the resulting AST.

Rule-specific disabling may be trickier as we would have to walk the sibling nodes until we encounter a *singlelineComment + newline*, then parse the comment to check if the rule is ignored.

kokarn:
I think we should go for // lesshint ignore:importantRule. It's a really
nice feature and I see no reason to go half-assed.

jwilsson:
I agree with @kokarn, we should definitely take the time to work this out properly since I think it's a real "must-have" feature.

shellscape:
Carried over from #57, declaring per-file options in a comment at the top of a file goes along with this. Since the other was closed, perhaps renaming this for both type of inline/infile options is appropriate.

shellscape:
@kokarn jshint syntax, which I think works really well in this case, follows their .jshintrc syntax. eg:

```
.somerule { background: url(...); } /* lesshint urlQuotes: false */
```

or

```
/* lesshint urlQuotes: false */
.somerule { background: url(...); } 
```
that makes configuring lesshint uniform wherever you choose to; .lesshintrc, inline, etc.

kokarn:
Yeah, that seems like a good way to do this. :+1:

shellscape:
Starting work on this one. Tracking on https://github.com/gilt/lesshint/tree/inline-rules

cageygee3:
Would love to see this in action as well. For our project, we always alphabetize our properties, but there are a couple cases where it makes sense to not do this:

- `display: none` comes first (usually with a comment explaining why)
- Browser-specific properties usually mess up the ordering of things

Any status update?

jwilsson:
[...]

PR #138, opened 2016-02-20
Add support for inline disabling/enabling of rules

jwilsson:
Here it is, the big one. The million dollar feature. Inline configuration!

This PR gives us the possibility to do these kind of things:
```less
// lesshint spaceBeforeBrace: false, emptyRule: false
.foo{ // Won't be reported
    color: red;
}

.bar {

} // Won't report this empty rule

// lesshint emptyRule: true
.baz{

} // But this will be reported, but the missing space before { won't

.foobar { // The next line won't be reported
    color:blue; // lesshint-line spaceAfterPropertyColon: false
}
```

As shown, all rules that aren't scoped to a single line are disabled until they're reenabled. At this time, rules that are disabled in `.lesshintrc` will stay disabled and I think that makes most sense.

I tried to make it as liberal as possible with regards to spaces. The only space that's required is the one after `lesshint` and `lesshint-line` (to prevent collisions) but other than that you're free to put spaces (or not) wherever you want, before `lesshint`, before `:` and so on.

I left the names open to give us the ability to add more configuration options to each rule in the future, for example changing `style` options inline (anything that's valid JSON is what I'm thinking really). This is the same way as ESLint does it.

But if anyone's got any better name suggestions, please share them!

Ref: #55


kokarn:
Looks good, great job dude! 
Only thing i have questions about is this
> At this time, rules that are disabled in .lesshintrc will stay disabled and I think that makes most sense.

What does that mean exactly?

jwilsson:
@kokarn It means that a rule that's disabled in `.lesshintrc` can't be enabled inline, for example:

.lesshintrc:
```
{
    ...
    stringQuotes: false
}
```

file.less:
```
// lesshint stringQuotes: true
.foo {
    background: url("image.png"); // Still won't report this
}
```

shellscape:
Ooo I have to disagree with that choice. That's counter-intuitive, and removes quite a bit of flexibility. There may be specific files that you want to enforce rules which have otherwise been disabled. That may be an edge case, but this happens to be a utility which caters to the edge cases, and the inline rulesets are specifically for even more granularity.

kokarn:
+1 for @shellscape 
Why would you not want to be able to enable a specific rule in one place only?

adam-lynch:
:+1: 

To confirm, if I want to disable something for two lines, I'd do this?

```less
.a {
    color: red;
    // lesshint aRule: false
    background: blue;
    outline: 1px solid green;
    // lesshint aRule: true
}

```

shellscape:
the more I look at that setup, the more it rubs on me. I'm a big big fan of the way that jshint/eslint does it. the inline config should really only affect the line after it. jshint allows you to disable the linters altogether for a range, but I'm not a fan of turning off a rule for one line and having to turn it back on on the line after. that creates a lot of unneeded cruft for what I imagine will be sparing use. 

sorry, not trying to be difficult on this one, but definitely have preferences.

jwilsson:
@adam-lynch Yes, exactly.

@shellscape, @kokarn You have a point in wanting to enable rules inline but I think it's a bit tricky right now (getting default options, , etc.). Perhaps we can leave it like this for now and keep iterating on it in later versions?

But I'm not sure I follow you, @shellscape, perhaps I gave a bad example in my initial comment. You can disable a single line by adding the comment at the end of it and that will only affect that line. If I'm reading the ESLint and JSHint docs correctly, that's how they do it too, albeit with a slightly different comment syntax but the usage is the same. Not trying anything here, just wanna make sure we understand each other :)

kokarn:
Ah, hmm. Should we really merge things just to improve on them later?

What I think shellscape is saying is that ESLint doesn't allow enabling for a whole block, only single lines. Correct?

shellscape:
This should clear up what I was trying to convey, and is what I'd recommend we follow: http://jshint.com/docs/#inline-configuration. It's highly effective. 

I'm in favor of iteration in 99% of situations. Since the feature is particularly limited by not being able to turn on linters inline, that will constitute a potentially breaking change and following semver will be a minor version. As long as that won't hold up completing the feature in the future, I'm good with it.

kokarn:
Suggestion: 
Only line-scopes, nothing for a block.
Maybe an exception for the whole file, but don't really like that. 
Might be needed for some specific files tho.

jwilsson:
@kokarn's suggestion is what we decided upon in #55 and I got carried away when starting to implement it. My bad!

I've added some of it, and just gonna finish up the last part of single line options before we're done. When I'm done, it'll be what we decided upon in #55 with the possibility to enable rules that are disabled in `.lesshintrc` as well.

kokarn:
Jay! Sounds like a great plan.

jwilsson:
After a lot of work and thinking I don't think it's possible to enable rules for a single line at this moment because of the way we're passing nodes to each linter. And quite frankly, I don't see the need for it either. I do, however, see the need to enable a disabled rule (or set other options for it) at the top of a file (which will be possible) but the single line option will only be able to disable it.

No other linter (JSHint, ESLint) allows it either. JSHint just uses a rule given earlier or falls back to defaults/a config file. And ESLint analyzes the whole file so a rule option given inline is applied everywhere (even if it occurs further down in the file).

But, to summarize:
* A comment at the top of a file will allow you to enable, disable, and configure linter(s) using the same format as in `.lesshintrc` (even a bit relaxed, not as strict as actual JSON).
* A trailing comment on a line will allow you to disable the specified linter(s) for that line.

Does this sound OK for everyone?

kokarn:
Yeah, sounds like a good way forward

miloszsobczak:
Actually this feature is something I was expecting when I first heard of this package, so for me - sounds great ;)

jwilsson:
If no one's got any objections I'll merge this tonight (CET) and tag a `1.4.0`.

jwilsson:
Merged in cf757cd9741cffeef966885de76168d50dc58716.

PR #143, opened 2016-02-22
teamcity fix

wareczek:
sorry... my bad :D square brackets are not allowed :)


jwilsson:
No worries, thanks for the quick fix!

PR #146, opened 2016-02-23
call reporter even if no errors found

wareczek:
TeamCity needs to know that test passed.

https://github.com/lesshint/lesshint/blob/master/lib/reporters/teamcity.js#L37

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -93,12 +93,6 @@ module.exports = function (program) {
 
         results = [].concat.apply([], results);
 
-        if (!results.length) {
-            exitDefer.resolve(EXIT_CODES.OK);
-
-            return;
-        }
-
         reporter = getReporter(program.reporter);
 
         if (reporter) {
@@ -110,6 +104,12 @@ module.exports = function (program) {
             }
         } else {
             console.error('Could not find reporter "%s".', program.reporter);
+        }
+
+        if (!results.length) {
+            exitDefer.resolve(EXIT_CODES.OK);
+
+            return;
         }
 
         hasError = results.some(function (result) {


jwilsson:
I'd like some additional feedback on this, do we think it can cause any BC changes? Ping @lesshint/owners @lesshint/committers

kokarn:
It shouldn't, right? 
Because it just calls the reporter which "should" only report what it got. 
Are we testing this properly?

wareczek:
Reporter "stylish" has forEach so empty errors result does nothing in output. TeamCity needs info to show "test passed" and this can't be exit code only.

Look on JSCS and JSHint reporters. They calls reporter without errors.

https://github.com/wurmr/jscs-teamcity-reporter/blob/master/teamcity-reporter.js
https://github.com/hongymagic/jshint-teamcity/blob/master/teamcity.js

Yes @kokarn, we have test for that in default reporter:
https://github.com/lesshint/lesshint/blob/master/test/specs/reporters/stylish.js#L65

shellscape:
Is it just me, or does it feel wrong to be modifying a core component of lesshint to suit a 3rd party service? I didn't see the teamcity reporter get added, but I would've had the same position as below I think.

I don't use lesshint through the cli, I use it through gulp, so admittedly there are some big discrepancies there. What I typically tend to find is that utilities such as these will have a base core reporter, and then allow add-on reporters; ala jshint-stylish fr jshint, nyan-reporter for mocha, etc. The core of those respective utils all operate the same way and the reporters each tweak how they run and the output.

With jshint, for example, we have a tertiary reporter which catches errors and tells gulp to exit with an error code if there are errors, and not just to treat them as warnings - which is the default behavior. 

All that being said, I don't think it's unreasonable for the CLI to call out to the reporter(s) at the end of each operation, regardless of whether or not there were errors/warnings. That will allow other reporters the opportunity to do whatever it is they want to do. But that shouldn't be limited to the cli. I don't think that we should have allowed the teamcity reporter to be added as a core component, rather it could have been an add-on project for lesshint, external to this repo.

kokarn:
I see your comment as 3 separate points @shellscape .

1. Should we modify the core to handle 3rd party stuff? 
I don't think that's a pleasant road, so let's avoid that.

2. Should we call the reporters regardless of error/warning? 
I think that's a good way to go. Should probably be up to the wrapper (cli/gulp/grunt/whatever) do decide what to do if it errors or not.

3. Should the teamcity reporter be included in this repo? 
That's the tough one I think. As i'm writing this i'm actually considering if we should move all reporters out of the main project and into their own projects under the lesshint org. Toughts @jwilsson and @wareczek ?

wareczek:
Ok, we can have reporters outside project. It's not a problem. But this pull request is about ability to call reporter and in my opinion this should be on reporter side to choose what to do. As I said before... JSCS and JSHint does that.

jwilsson:
Alright, here's what I'm thinking and some reasoning.

When I originally added the reporter support, I looked at JSHint, ESLint, etc. and saw that they include a bunch of reporters. At the time I thought `lesshint` could do the same thing but wanted the community to add the reporters needed (`stylish` is enough for me). But in the spirit of modularization I absolutely agree that we should move the reporters outside into separate repos under the lesshint organization and let people require the one(s) they need themselves. We should be able to do this now as long as we add `stylish` as a dependency to lesshint (which can be removed in `2.0`). Then we can make it easier for our Grunt and Gulp plugins to utilize these reporters as well (I started some work in the Gulp plugin but it can definitely be improved and expanded upon).

I also think that this PR is a good change, and if no one sees any backwards incompatible changes with it I suggest we merge it.

And, since we haven't published anything with the TeamCity reporter yet, we are still able to move this to a separate repo under the lesshint organization with @wareczek in charge (if you're interested of course).

shellscape:
:+1:  solid plan I think

kokarn:
SGTM :+1:

wareczek:
Ok, we can do that way. How should work this plugin and repo would look? Could you provide any example @jwilsson ? :)

jwilsson:
@wareczek I'll setup a new repo, move the files there, and add you to it later tonight.

miloszsobczak:
Absolutly agree with the modular architecture. I'd love to code new reporter as a modular one.

jwilsson:
@wareczek I've added you to the https://github.com/lesshint/reporter-teamcity repo and I'm merging this now. I'll also remove the TeamCity things from this repo.

Thanks for the patch :)

wareczek:
ok :) thx for repo

PR #148, opened 2016-02-24
spaceAroundComma & spaceBetweenParents - multiline friendly

### LINKED ISSUE #119: spaceAroundComma linter breaks on multiline definition
For very long definitions, as you might find with gradients, we'd like to be able separate each step on to a new line without the `spaceAroundComma` linter flagging it. Perhaps this needs to be an additional option?

Something like the following should be considered valid, in my opinion:
```less
.thing {
    background: linear-gradient(
        to bottom,
        transparent 0,
        transparent 25%,
        red 25%,
        red 75%,
        transparent 75%,
        transparent 0
    );
}
```

### LINKED ISSUE #130: spaceBetweenParents isn't multi-line friendly
This fails:

```less
.button-base(
  @baseBackgroundColour,
  @focusBackgroundColor: @baseBackgroundColour,
  @hoverBackgroundColour: lighten(@baseBackgroundColour, @buttonBackgroundDelta),
  @activeBackgroundColour: darken(@baseBackgroundColour, @buttonBackgroundDelta)
) {
  // ...
}
```

wareczek:
fixes #119 & fixes #130

Extra checking other side
`spaceAroundComma` with option `after` or `before` were positive on `rgb(100 , 100 , 100)`

Fixed reporting of column number with prev elements. Prev should report `end.column` not `start.column`.

### DIFF
--- a/lib/linters/space_around_comma.js
+++ b/lib/linters/space_around_comma.js
@@ -12,64 +12,76 @@ module.exports = {
         var self = this;
 
         node.forEach('operator', function (element, index) {
-            var startElement;
-            var nextElement;
-            var prevElement;
-            var message;
+            var next = node.content[index + 1];
+            var prev = node.content[index - 1];
 
             if (element.content !== ',') {
                 return;
             }
 
-            nextElement = node.content[index + 1];
-            prevElement = node.content[index - 1];
-
-            // setting `startElement` equal to nextElement as the default for
-            // `style`is `after`
-            startElement = nextElement;
-
             switch (config.style) {
                 case 'after':
-                    if (nextElement.type !== 'space' || nextElement.content !== ' ') {
-                        message = util.format(self.message, '', 'followed', 'one');
+                    if (next.type !== 'space' || next.content !== ' ' && next.content.indexOf('\n') !== 0) {
+                        results.push({
+                            column: next.start.column,
+                            line: next.start.line,
+                            message: util.format(self.message, '', 'followed', 'one')
+                        });
                     }
-                    break;
 
+                    if (prev.type === 'space' && prev.content === ' ') {
+                        results.push({
+                            column: prev.end.column,
+                            line: prev.end.line,
+                            message: util.format(self.message, ' not', 'preceded', 'any')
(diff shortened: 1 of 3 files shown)


kokarn:
Correct me if i'm wrong, but this will always allow newlines?
What we discussed in #119 and #130 is that there needs to be an option on what should be allowed and what shouldn't.

wareczek:
Yes, newlines are allways allowed now. Do we really need this extra option? 

Your example from #119 (https://github.com/lesshint/lesshint/issues/119#issuecomment-177874392) is not acceptable for that rule, because they are only for function calls already (https://github.com/lesshint/lesshint/blob/master/lib/linters/space_around_comma.js#L7).

So in my opinion newline should not be treated like space. If you have small line limit and couple of nesting, you will have to quit option `after` and activate `after-newline`. Not quite nice way :(

jwilsson:
Hmm, I don't know actually. But I guess @cvisco's suggestion 
> This is potentially naive solution, but I wonder if just allowing newlines after—or before—commas inside ( or ) would work?

(ref: https://github.com/lesshint/lesshint/issues/119#issuecomment-178040665) makes the most sense since it's only in functions you would want to check it.

wareczek:
`after` is now like @cvisco's suggestion, but what with `before`? Code like:
```css
.thing {
    background: linear-gradient(
        to bottom ,
        transparent 0 ,
        transparent 25% ,
        red 25% ,
        red 75% ,
        transparent 75% ,
        transparent 0
    );
}
```
should be valid? Looks wierd :D

kokarn:
@jwilsson I'll let you handle this until I'm back from vacation (next week)

shellscape:
@kokarn @jwilsson this one has gone zombie mode. let's not let it wander for eternity

shellscape:
@wareczek no longer relevant, or? wouldn't want to lose a solid contribution if we can still use it!

wareczek:
#130 you've fixed and #119 i'll try from ground up on new branch ;)

shellscape:
awesome, thanks!

PR #150, opened 2016-03-03
fix(decimalZero): correct false positive

### LINKED ISSUE #137: False Positive with "decimalZero"
Given `decimalZero: "leading"` and the following Less code...

```less
.foobar {
  color: rgba(255, 255, 255, .5);
}
```

I should get an error on line 2, because `.5` should be `0.5`

CITguy:
* check all "number" values in declaration
* more robust regexes
* refactored tests

closes #137

### DIFF
--- a/lib/linters/decimal_zero.js
+++ b/lib/linters/decimal_zero.js
@@ -9,58 +9,67 @@ module.exports = {
 
     lint: function decimalZeroLinter (config, node) {
         var number;
-        var value;
-        var output = {
-            inclusion: 'with',
-            type: null
-        };
+        var result = [];
+        var self = this;
 
-        node.forEach('value', function (element) {
-            value = element.first('dimension');
-            value = value && value.first('number');
-        });
+        node.traverseByType('number', function (element) {
+            var output;
 
-        // Bail if NOT floating point format or no value
-        if (!value || !(/^-?(\d*\.\d*)/.test(value.content))) {
-            return;
-        }
+            number = element.content;
 
-        number = value.content;
+            /*
+             * Bail if:
+             *  - Not a floating point number
+             *  - Float parsed as 0 (e.g. 0.0, .0, 0.)
+             */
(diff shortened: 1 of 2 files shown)


lib/linters/decimal_zero.js
```diff
@@ -9,58 +9,67 @@ module.exports = {
...
-            inclusion: 'with',
-            type: null
-        };
+        var result = [];
+        var self = this;
 
-        node.forEach('value', function (element) {
-            value = element.first('dimension');
-            value = value && value.first('number');
-        });
```
CITguy: This logic would iterate over all "value" nodes, but the `value` variable would only be set to the last one in the iteration.

lib/linters/decimal_zero.js
```diff
@@ -9,58 +9,67 @@ module.exports = {
...
+        node.traverseByType('number', function (element) {
```
CITguy: Wrapped linting logic with `traverseByType()` to run linting on *each* number found in declaration.

jwilsson:
LGTM :+1:

CITguy:
Is AppVeyor having issues? I don't think I have access to see the lesshint project on appveyor.

CITguy:
reopened to re-trigger app integrations

jwilsson:
I'm gonna delete the project from AppVeyor and add it again. It's pointing to the old project (jwilsson/lesshint) so that might cause some issues.

jwilsson:
Reopened to try and trigger new build. Sorry for spamming notifications.

CITguy:
no problem :)

jwilsson:
I can't seem to trigger a AppVeyor build of this but Travis passes and it seems unlikely that AppVeyor will fail, and if it does, we'll catch it anyway.

I'm good with merging unless someone else thinks different? :)

CITguy:
Travis tested it 7 times, and passed each time. I'm good with merging, if you are.

Issue #152, opened 2016-03-15
zeroUnit is incorrectly applied to opacity


ajaswa:
According to the spec, the `opacity` property does not have a unit for its value. When running lesshint against a less file with `opacity` properties, the `Unit should not be omitted on zero values.` warning comes up.

Reference docs:
https://www.w3.org/wiki/CSS/Properties/opacity
https://www.w3.org/TR/css3-color/#transparency
https://developer.mozilla.org/en-US/docs/Web/CSS/opacity

jwilsson:
Confirmed. I'm guessing from our Twitter conversation that you're interested in patching it?

I can't come up with any other properties which works like `opacity` right now, but if you do please add them here too!

ajaswa:
Indeed, I'd like to take a stab at it. I can't come up with anything else either. If I do I'll include them here.

ajaswa:
z-index also has a non-unit numerical value.

https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

kokarn:
How about other things? 
It's recommended to use line-height without a unit for better scaling, but it's possible to use with units.

shellscape:
Sounds like there needs to be a property exclusion list for this one.

ajaswa:
@shellscape indeed. I added an array of properties that can be easily added to when more properties are "discovered" that have non-unit values. Though I'm not sure I've done it in the right place but it seems to work.

Right now the list includes `opacity`, `z-index' and `line-height`.

jwilsson:
@shellscape were you thinking a configurable list or a hardcoded one? I'm fine with either.

shellscape:
following previous patterns, a configurable list with defaults seems to be a good idea.

ajaswa:
I disagree. Allowing users/devs to configure a list of properties to determine what the CSS spec has already defined seems like a very bad idea. If thats not the case then why not have all the units configureable?

shellscape:
@ajaswa lesshint linters follow a pattern of configurability. I don't quite get what "all the units" means in this context, but allowing users to add properties to that list the spec already defines is certainly reasonable.

ajaswa:
What I mean by "all the units" is that by your logic this line should also be set in configuration: https://github.com/lesshint/lesshint/blob/master/lib/linters/zero_unit.js#L12

shellscape:
[...]

PR #153, opened 2016-03-15
Skip zeroUnit when the property is supposed to have a non-unit number.

ajaswa:
@jwilsson This should take care of #152, let me know if you have any feedback.

### DIFF
--- a/lib/linters/zero_unit.js
+++ b/lib/linters/zero_unit.js
@@ -14,6 +14,13 @@ module.exports = {
         var number;
         var value;
         var unit;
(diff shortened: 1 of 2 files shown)


jwilsson:
Could you just squash the commits and I'm good. Don't know if anyone else's got any comments.

I think we should hold of with the merge until we're able to run the tests reliably (due to a ESLint bug, https://github.com/eslint/eslint/issues/5586).

shellscape:
@ajaswa why was your fork deleted?

PR #156, opened 2016-03-18
brings zeroUnit inline with the spec, provides config for both units and property names.

shellscape:
Not sure why #153 was closed and the fork was deleted, but this PR contains the same work done by @ajaswa, with the added option of allowing users to add units and excluded properties to the existing list.

### DIFF
--- a/lib/linters/zero_unit.js
+++ b/lib/linters/zero_unit.js
@@ -14,6 +14,23 @@ module.exports = {
         var number;
         var value;
         var unit;
+        var ignoredProperties = ['opacity', 'z-index'];
+        var property = node.first('property').first('ident').content;
+
+        if (config) {
+            if (config.ignore && config.ignore.length) {
+                ignoredProperties.push(config.ignore);
+            }
+
+            if (config.units && config.units.length) {
+                units.push(config.units);
+            }
+        }
+
+        // The property shouldn't be checked for units
+        if (property && ignoredProperties.indexOf(property) !== -1) {
+            return;
+        }
 
         node.forEach('value', function (element) {
             value = element.first('dimension');
@@ -30,12 +47,12 @@ module.exports = {
 
         // Nothing to lint found, bail
         if (!number || parseFloat(number.content) !== 0) {
-            return null;
+            return;
         }
 
         // Unit is required, nothing to do
         if (unit && units.indexOf(unit.content) === -1) {
-            return null;
+            return;
         }
 
         number = number.content;
@@ -54,9 +71,7 @@ module.exports = {
 
                 break;
             default:
-                throw new Error(
-                    'Invalid setting value for zeroUnit: ' + config.style
-                );
+                throw new Error('Invalid setting value for zeroUnit: ' + config.style);
         }
 
         if (!valid) {
--- a/lib/linters/README.md
+++ b/lib/linters/README.md
@@ -800,9 +800,11 @@ The [CSS spec](http://dev.w3.org/csswg/css-values/#url-value) also recommends th
 ## zeroUnit
 Zero values should include a unit for consistency with other values.
 
-Option     | Description
----------- | ----------
-`style`    | `no_unit`, `keep_unit` (**default**)
+Option      | Description
+------------| ----------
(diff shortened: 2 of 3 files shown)


shellscape:
We should retrigger the CI jobs after #157 is merged. oodles of indentation issues with the linter.

lib/linters/zero_unit.js
```diff
@@ -14,6 +14,23 @@ module.exports = {
         var number;
         var value;
         var unit;
+        var excludeProperties = ['opacity', 'z-index', 'line-height'];
+        var property = node.first('property').first('ident').content;
+
+        if (config) {
+            if (config.excludeProperties && config.excludeProperties.length) {
+                excludeProperties.push(config.excludeProperties);
```
jwilsson: Isn't `concat` better than `push` here and below for the units?

jwilsson:
Do you think we need an extra test for `excludeProperties` and/or `units`?

lib/linters/zero_unit.js
shellscape: we'd only want to use concat if the intention was to return a new array as the result. since excludeProperties is just a reference point, I didn't see a need for that, and figured it could be modified in place safely.

shellscape:
if you think it's necessary, I have no objection. it was such straightforward code that I skipped it. call me lazy :)

shellscape:
copying from @jwilsson's comment on #152...

1. In zeroUnit, we keep a list of properties defined without values in the CSS spec (i.e. opacity and z-index). And add a exclude option which is merged with the default, hardcoded ones. So no one's able to remove things in the CSS spec but add something if, for example someone wants to exclude padding for whatever reason.
2. I don't think line-height needs to be included in the defaults, though. It's a bit too opinonated and frankly, line-height: 0 doesn't make much sense to me ;)
3. I don't know what my opinion on excluding properties is, though. Perhaps we should keep that discussion in another issue?

----

1. agreed. essentially what we've setup, expanded upon the original PR
2. no problem here with removing that from the default list
3. I think it makes sense, given the context. we really don't want to be throwing false positives if we can avoid it.

Also, I like "ignore" better than "exclude" - thoughts?

ajaswa:
I fundamentally disagree with this direction and as such I can't support something I disagree with which is why I've pulled my PR and fork.

The CSS spec moves at a glacial pace, so to allow users to add their own properties, units and what not is unfounded for the sake of keeping "inline with the spec".

That said what you've done is taken my simple bug fix and turn it into a feature and refactor of how properties/units are set in lesshint linters. And the reasoning for this is unfounded as well, since there is only one other linter (borderZero) that has hardcoded properties or units, so there is no pattern as you suggest to follow. As a developer I hope you can appreciate the issues with scope creep.

While purposing new features and refactors of how things work isn't a bad thing, but you've stopped with just one of the linters, which now makes this feature incomplete.

Please remove my code as I do not want to add my support to this direction.

shellscape:
@ajaswa the code in your closed PR has been refactored, as it's too common to outright remove the exact patterns, and all mention of your original PR has been scrubbed from the commit.

disagreements will happen in open source, and taking your ball and going home will pave a rough path to contribution in future projects. I have disagreed with many maintainers over the years, but have been gracious that my contributions have improved a project, even if they were purposed beyond what I intended.

shellscape:
@jwilsson #1 and #2 from your list are done. and extra specs added for the config.

jwilsson:
This is merged in 728f37a73d117d6352d2bf146fa09240be4453a2.

@shellscape I changed these things:
* Changed `ignore` back to `exclude` (for consistency).
* Changed `push` to `concat`. The reasoning behind this was to transparently handle strings and arrays so we'll always end up with single dimension arrays.
* Added a new check for `units`. I noticed a test failure when I changed to `concat`, the old `units` array contains the units where a zero value doesn't require a unit since angles does, for example. It now does its own check for excluded units. Hope this sentence makes sense :)

If you have any thoughts or changes let me know or just push it. If not, I'll tag a new version tonight or tomorrow (CET).

shellscape:
:+1:  all good with me.

PR #157, opened 2016-03-18
synchronous gulp task execution, linter indentation fixes

shellscape:
the linter was pitching a fit about array definition indentation when the array def contained multiple objects.

also separated the lint and config task deps for the test task, enforcing a linear and synchronous execution. was having an issue of the coverage task failing because of a syntax error before the lint task could tell me where it was.


jwilsson:
Great, thanks for tackling it. I filed an issue at the ESLint repo but don't know if it's going to be fixed.

PR #160, opened 2016-03-21
Fix possible decimalZero regression with floats greater than 1

jwilsson:
I think I found a minor regression in the recent patch for `decimalZero`. 

The issues I found:
* Earlier `1.25` wouldn't be reported when `style` was `trailing` but now it was. 
* Earlier `1.250` wouldn't be reported when `style` was `both` but now it was. 

@CITguy Could you just take a look at this fix and see if it does what you expect?

### DIFF
--- a/lib/linters/decimal_zero.js
+++ b/lib/linters/decimal_zero.js
@@ -33,7 +33,7 @@ module.exports = {
 
             switch (config.style) {
                 case 'leading':
-                    if (!/^-?(0\d*)\.(\d*[1-9])$/.test(number)) {
+                    if (!/^-?(\d+)\.(\d*[1-9])$/.test(number)) {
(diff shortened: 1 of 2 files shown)


jwilsson:
I'm merging this so we can tag a new release. All new and old tests pass.

CITguy:
Which is the purpose of `decimalZero`?

1. to ensure that we don't have a leading/trailing decimal (`.5` vs `0.5`, `5.` vs `5.0`)?
2. to enforce a literal leading/trailing 0 in all cases (regardless of digits immediately adjacent to decimal (`1.35` vs `01.350`)?
3. other?

jwilsson:
@CITguy I opened a new issue (#165) to continue the discussion.

PR #161, opened 2016-03-22
Fix fragment identifier (#depthlevel)

monoblaine:
The fragment identifier for the **depthLevel** section should be `#depthlevel` (lowercase).


jwilsson:
Thanks!

PR #167, opened 2016-03-30
Fix gonazel-pe not being fetched when git protocol port is blocked

abhishekdev:
- Provides a workaround within enterprises where the default git protocol port is blocked.
- Use github URL format in package.json instead of a git only URL. This prevents npm from failing when and rather use https ports instead of git protocol port (9418) id git global config are set via: `git config --global url."https://".insteadOf git://`

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "chalk": "^1.0.0",
     "commander": "^2.8.0",
     "exit": "^0.1.2",
-    "gonzales-pe": "git://github.com/gilt/gonzales-pe.git#dev",
+    "gonzales-pe": "gilt/gonzales-pe#dev",
     "lodash.findindex": "^4.0.1",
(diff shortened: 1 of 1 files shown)


abhishekdev:
The npm github URL pattern support was introduced in `v1.1.65` <sup>*[source](https://docs.npmjs.com/files/package.json#github-urls)*</sup>

Since node `0.10.x` shipped with npm `v1.2.14` this change should be fine for backward compatibility.<sup>[source](https://nodejs.org/en/blog/release/v0.10.0/#2013-03-11-version-0-10-0-stable)</sup>

jwilsson:
Nice catch!

And thanks for the detailed explanation and patch :smile:

jwilsson:
Released in `1.5.2`, let us know if it doesn't work.

PR #169, opened 2016-04-07
Strip trailing slashes from directory argument

jasonewang:
-Fix case where report fullPath field would have multiple slashes
separating directories

ex. `lesshint src/` would produce reports with fullPaths like `src//foo/bar/baz.less`


jwilsson:
Thank you! 🎉

PR #171, opened 2016-04-20
Reporter changes

### LINKED ISSUE #147: Move stylish reporter to its own package
Just to make sure we don't forget our discussion in #146.

The `stylish` reporter should be moved to its own repo and package. What we need to do right now:
- [x]  Move files to the new repo.
- [x]  Remove the `chalk` dependency from `package.json`.
- [x]  Add a special case for `stylish` when we load reporters so it'll still work with the old name (since the new one will be `lesshint-reporter-stylish` or similar).
- [x]  Add deprecation notice when trying to use `stylish` with the old name.
- [x]  Some minor doc updates.
- [x] Add a general `getReporter()` method to allow other packages to use our loading logic.

EDIT:
Added `getReporter()` point.

jwilsson:
Closes #147.

The `stylish` reporter now lives in https://github.com/lesshint/reporter-stylish and there's a simple, no colors reporter included here which is used when no reporter is passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,30 +133,46 @@ Exit status code   | Description
 These codes were chosen with regards to the [preferable exit codes](http://www.gsp.com/cgi-bin/man.cgi?section=3&topic=sysexits).
 
 ## Reporters
-As of `0.8.0` the ability to specify custom reporters has been added. These can do anything from just printing something to the terminal to generate custom reports.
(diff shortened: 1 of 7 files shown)


shellscape:
👍  all looks good. my only teensy suggestion would be to name the stylish reporter "lesshint-stylish" following jshint-stylish. but that is of course, a matter of personal preference.

Issue #177, opened 2016-05-19
propertyOrdering and singleLinePerProperty bugs


wareczek:
```
.form-group {
    margin-bottom: 0;

    .form-control {
        height: auto;
    }
}
```

> Warning: bootstrap.less: line 5, col 9, propertyOrdering: Property ordering is not alphabetized
> Warning: bootstrap.less: line 5, col 9, singleLinePerProperty: Each property should be on its own line.

```
.form-group {
    margin-bottom: 0;
}

.form-control {
    height: auto;
}
```

> Warning: bootstrap.less: line 2, col 5, singleLinePerProperty: Each property should be on its own line.
> Warning: bootstrap.less: line 6, col 5, singleLinePerProperty: Each property should be on its own line.

shellscape:
what version are you using?

wareczek:
As in title ;) 2.0 rc

jwilsson:
I got a fix for the `propertyOrdering` issue in [bug-2.0/propertyOrdering](https://github.com/lesshint/lesshint/commits/bug-2.0/propertyOrdering).

wareczek:
fix for `singleLinePerProperty` bug:

`if (!next || (next && ['comment', 'rule'].indexOf(next.type))) {`

on line:
https://github.com/lesshint/lesshint/blob/master/lib/linters/single_line_per_property.js#L6

@jwilsson correct? ;) sorry for no PR, i'm busy now :(

jwilsson:
Fully closed with 943f7c1.

PR #178, opened 2016-05-20
2.0 propertyOrdering bugs

jwilsson:
This fixes two issues:
* It treated all the declarations in nested rules as one and thus complained about them. See #177.
* It would complain on variables that weren't alphabetically ordered.

### DIFF
--- a/lib/linters/property_ordering.js
+++ b/lib/linters/property_ordering.js
@@ -25,7 +25,8 @@ module.exports = {
 
             property = child.prop;
 
-            if (!property) {
+            // Ignore declarations without a property and variables
+            if (!property || (property && /^@/.test(property))) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


lib/linters/property_ordering.js
```diff
@@ -15,17 +15,18 @@ module.exports = {
             throw new Error('Invalid setting value for propertyOrdering: ' + config.style);
         }
 
-        node.walkDecls(function (declaration) {
+        node.each(function (child) {
```
shellscape: so why wasn't walkDecls working correctly here? was it because it was cascading into the declarations of nested stuffs?

shellscape:
👍

wareczek:
Works great :)

lib/linters/property_ordering.js
jwilsson: Yeah, so this:

```less
.form-group {
    margin-bottom: 0;

    .form-control {
        height: auto;
    }
}
```

would be treated as 
```less
.form-group {
    margin-bottom: 0;
    height: auto;
}
```

PR #180, opened 2016-05-22
Fixes to column reporting

jwilsson:
Fixed a bunch of issues where the column reporting was a bit off or wrong.

I also found a false positive in `spaceAfterPropertyColon` where it would report errors if there were spaces before the colon.


jwilsson:
We're probably gonna wanna a helper for this, lots of duplication right now :/

shellscape:
👍 

and yes to a helper. might be able to tackle that next week.

jwilsson:
I say we merge this and look at a helper later.

Issue #165, opened 2016-03-29
decimalZero behavior change


jwilsson:
Carrying over from @CITguy's comment in https://github.com/lesshint/lesshint/pull/160#issuecomment-201007411:

> Which is the purpose of decimalZero?
> 
> 1. to ensure that we don't have a leading/trailing decimal (.5 vs 0.5, 5. vs 5.0)?
> 2. to enforce a literal leading/trailing 0 in all cases, regardless of digits immediately adjacent to decimal (1.35 vs 01.350)?
> 3. other?

My two cents are for no. 1 but I'm not attached to anything. If we want to change the behavior, now would be a perfect time since I'm thinking we'll tag a `2.0` when we're done with the PostCSS conversion.

CITguy:
I'm fine with option 1. I just wanted to make sure that we have that behavior documented and ensure that the linter is tested as such. I believe #160 corrects the leading zero portion of this behavior, so we'd need another PR to correct the trailing zero behavior.

PR #188, opened 2016-06-07
Behavior change in decimalZero 

jwilsson:
As per #165.

TLDR, `decimalZero` will now only check if there's always/never a leading/trailing decimal number, not that it's always a zero there (even if there were other numbers too).

### DIFF
--- a/lib/linters/README.md
+++ b/lib/linters/README.md
@@ -119,23 +119,18 @@ Option     | Description
 ### invalid
 ```less
 .foo {
-  font-size: 1.0em;  // leading
-  font-size:  .5em;  // leading, trailing, both
-  font-size: 0.5em;  // trailing, both, none
-  font-size: 1.5em;  // trailing, both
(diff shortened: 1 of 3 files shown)


jwilsson:
CI build: https://ci.appveyor.com/project/lesshint/lesshint/build/11

shellscape:
is the linter's name still correct given that it doesn't care about zeros anymore?

jwilsson:
I think I explained poorly, it still cares about zeros but only if there's no other numbers before/after the decimal comma.

```less
font-size: .5em; // Will report this with style "leading" and "both" but not with style "none"
font-size: 1em; // Will report this with style "trailing" and "both" but not with style "none"
font-size: 1.5em; // Will never report this
```

Previously this would happen:
```less
font-size: 1.5em; // Would be reported with style "trailing" and "both"
font-size: 1.0em; // Would be reported with style "leading"
```

shellscape:
ah very solid change then

PR #189, opened 2016-06-08
Don't check function arguments in zeroUnit

### LINKED ISSUE #184: zeroUnit erroneously report zeros in color functions
In `2.0-rc` `zeroUnit` will report this:
```css
color: rgb(0, 0, 0);
```

I've been thinking and I don't think we can just exclude certain functions since there are a lot of them in Less, Math and Color for example (full reference: http://lesscss.org/functions/).

I suggest we just ignore all function calls in `zeroUnit`.

jwilsson:
Closes #184

### DIFF
--- a/lib/linters/zero_unit.js
+++ b/lib/linters/zero_unit.js
@@ -35,8 +35,14 @@ module.exports = {
         ast = parser(node.value).parse();
 
         ast.walk(function (child) {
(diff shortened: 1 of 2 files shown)


jwilsson:
Updated with a better check which just looks at the first child level (doesn't traverse the whole tree, just like in `1.x`).

PR #190, opened 2016-06-08
Ensure singleLinePerProperty checks each rule on its own

jwilsson:
Props @wareczek (idea taken from an old comment that was deleted for some reason).

Ref #177.

### DIFF
--- a/lib/linters/single_line_per_property.js
+++ b/lib/linters/single_line_per_property.js
@@ -2,8 +2,9 @@
 
 var isValidAfter = function (node, index) {
     var next = node.parent.nodes[index + 1];
+    var types = ['comment', 'rule'];
 
-    if (!next || (next && next.type === 'comment')) {
+    if (!next || (next && types.indexOf(next.type) !== -1)) {
         // Only valid if followed by comment
         return true;
     }
--- a/lib/linters/single_line_per_property.js
+++ b/lib/linters/single_line_per_property.js
@@ -1,11 +1,19 @@
 'use strict';
 
+var hasNewlineBefore = function (node) {
+    return /\n/.test(node.raws.before);
+};
+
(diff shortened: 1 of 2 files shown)


wareczek:
I figured out that this idea was buggy is some case... but now don't remember it :D

lib/linters/single_line_per_property.js
```diff
@@ -2,8 +2,9 @@
 
 var isValidAfter = function (node, index) {
     var next = node.parent.nodes[index + 1];
+    var types = ['comment', 'rule'];
 
-    if (!next || (next && next.type === 'comment')) {
+    if (!next || (next && types.indexOf(next.type) !== -1)) {
```
shellscape: would Array.includes be faster here?

wareczek: it's ES6 feature :P

shellscape: d'oh!

wareczek:
Fails on:
```css
.modal-content {
    font-weight: normal;
    height: 20px;

    .close {
        background-color: red;
    }
}
```

jwilsson:
Aha, thanks @wareczek!

Do you wanna take another look at it?

jwilsson:
I have a patch ready, just wanted to get your opinion on it, the difference from now is that it'll emit two errors for these kind of lines (one for each property):

```less
.modal-content {
  font-weight: normal; height: 20px;
}
```

I think that's OK. If everyone agrees, I'll update the tests to reflect it (since some of them expect one error now where there will be emitted two instead).

wareczek:
It's ok :)

shellscape:
@jwilsson you've been kicking ass lately!

jwilsson:
@shellscape I wanna get `2.0` out the door ;)

PR #197, opened 2016-06-19
Added `linters` option to config

### LINKED ISSUE #101: Linter plugin support
I was thinking about some sort of plugin support. That users of `lesshint` can create their own linters which doesn't necessarily belong in core (company specific linters or whatever) but still want to use the infrastructure provided by `lesshint` to lint their code.

We already got most of the structure in place for this in `linter.js`. What we need to do is basically this (probably forgot something):
* Merge our current list of linters with a `plugins` array from `.lesshintrc`.
* Write some docs on what a linter needs to do. Expose a `lint` method, return an array etc.

JoshuaKGoldberg:
Allows the user to provide an array of paths to custom linters that will be added to the list of default linters.

Holding off on adding docs until a spec for it is agreed on.

Current proposal:
```json
linters: ["../test/plugins/sample"]
```

Fixes #101.


jwilsson:
This looks great!

I'm fine with the name `linters` instead of `plugins`, but lets get some other opinions on it too.

One thing I'm thinking about though is whether it should be possible to load linters from both a path but also a module, like we do in [`getReporter()`](https://github.com/lesshint/lesshint/blob/943f7c156c8f8571f52c09bb00666b60bd5a2287/lib/lesshint.js#L120)?

I also think we should add a CLI flag `-p`/`--plugins` (`-l`/`--linters`) while we're at it too.

lib/linter.js
```diff
@@ -174,6 +178,22 @@ var runChecks = function (ast, fullPath, lines, config) {
...
+    }
+
+    if (!(config.linters instanceof Array)) {
+        throw new Error('linters should be an array of file paths');
+    }
+
+    return defaultLinters.concat(
+        config.linters.map(
+            function (linterPath) {
+                return require(linterPath);
```
shellscape: this assumes a string. I'd like to see support for already-required modules (eg. Objects) here. 

```js
linters: [
  require('./lib/my-fancy-linter')(someConfig);
]
```

That'd negate the need for another 'linterConfig' section for configuring dynamic linters at load-time. I know we don't really do that now, but something like the proposed browser prefix linter could potentially use multiple source, and then we'd be mixing load-config with linter-config, or stuck creating another section. that smells of Grunt type config, and I'd like to avoid that. instead, following on the path of modules like babel, supporting the already-required modules.

(it's early here, let me know if that doesn't make sense :))

shellscape:
I'm good with `linters`

Echoed @jwilsson 's position on modules and strings, before I saw his comment.

CLI flag is a good catch, should def be in there.

Great start to this feature

JoshuaKGoldberg:
+1 for adding a CLI flag. Would you prefer it be in this PR or a followup one?

Re loading linters from paths, I like the idea of letting either paths or already-required modules being permitted:

```js
linters: [
  './lib/my-plain-linter',
  require('./lib/my-fancy-linter')(someConfig)
]
```

@shellscape does this go against what you're saying? I'm not familiar with what you're doing with runtime loading unfortunately.

shellscape:
actually that's exactly what I was suggesting haha.

lib/linter.js
```diff
@@ -178,20 +178,24 @@ var runChecks = function (ast, fullPath, lines, config) {
     return results;
 };
 
-var mergeLinters = function (defaultLinters, config) {
+var mergeLintersFromConfig = function (defaultLinters, config) {
```
shellscape: any reason for the fn name change? thought it was pretty clear before what the fn did, so I'm curious :)

JoshuaKGoldberg: Ehh it seemed a little inaccurate to me. It wasn't merging two linter arrays; it was merging a linter array with linters and/or linter descriptions from a config.

JoshuaKGoldberg:
> Coverage increased (+0.02%) to 98.891%

Love it.

test/specs/cli.js
```diff
@@ -140,6 +140,55 @@ describe('cli', function () {
...
+
+        result = cli({
+            args: [path.dirname(__dirname) + '/data/files/file.less'],
+            config: path.resolve(process.cwd() + '/test/data/config/linters.json'),
+            linters: ['../test/plugins/otherSampleLinter']
+        });
+
+        return result.fail(function (status) {
+            console.log.restore();
+            expect(status).to.equal(1);
```
JoshuaKGoldberg: How would you prefer this be tested? Theoretically if the feature were broken and only one of the linters were included in the test, it wouldn't catch it.

test/specs/cli.js
```diff
@@ -203,9 +252,8 @@ describe('cli', function () {
         });
 
         return result.fail(function (status) {
-            expect(status).to.equal(2);
-
             console.log.restore();
+            expect(status).to.equal(2);
```
JoshuaKGoldberg: *(irrelevant change: lets the `.restore()` call happen if `status` isn't 2)*

test/specs/cli.js
jwilsson: Hmm, I'm not sure actually. Your thoughts @shellscape?

jwilsson:
Great job with the CLI flag btw :)

JoshuaKGoldberg:
Added docs to README. I think the docs in https://github.com/lesshint/lesshint/issues/101#issuecomment-226110767 are better left to you folks. What I wrote is basically copypasta of existing stuff.

JoshuaKGoldberg:
@shellscape @jwilsson bump?

shellscape:
@JoshuaKGoldberg i can take a peek tonight

test/specs/cli.js
shellscape: I'd say we should add a test linter which fails regardless of the less passed in, and then check the response from that. how's that sound?

shellscape:
I'm gonna merge this one. great commit, and the added tests are a big bonus.

JoshuaKGoldberg:
Woot!

test/specs/cli.js
JoshuaKGoldberg: That's a good idea. I can do that.

PR #212, opened 2016-07-13
fixes #119: adds allowNewline option to spaceAroundComma

### LINKED ISSUE #119: spaceAroundComma linter breaks on multiline definition

shellscape:
This is my proposal for the spaceAroundComma fix when newlines are used to separate comma-delimited lists. `allowNewline` set to `true` will skip any verification of space around a comma if a newline is detected, based on the `style` option chosen.

### DIFF
--- a/lib/linters/qualifying_element.js
+++ b/lib/linters/qualifying_element.js
@@ -26,7 +26,7 @@ module.exports = {
             }
 
             result.endsWith = selector.last.type;
-
+            console.log(selector.nodes);
             if (!result.hasTag) {
                 selector.nodes.forEach(function (element) {
                     if (element.type === 'tag') {
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -132,6 +132,7 @@
     },
 
     "spaceAroundComma": {
+        "allowNewline": false,
         "enabled": true,
         "style": "after"
     },
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -125,8 +125,6 @@ var runChecks = function (ast, fullPath, lines, config) {
         linters.forEach(function (linter) {
             var options = config[linter.name];
             var lint;
-            var matches;
-            var pseudoParams;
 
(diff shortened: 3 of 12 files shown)


wareczek:
nice one ;) i like `allowNewline` as an option

jwilsson:
👍 

The tests are passing but branch coverage dropped below 95%, that's why the build's failing. @shellscape I'll leave it to your discretion.

shellscape:
@jwilsson any ideas on how to remedy that? I'd rather not let that go without giving a try at fixing :)

jwilsson:
@shellscape Hmm, when looking at the HTML report generated when running `npm test` locally it's actually `qualifyingElement` and `importantRule` that's the real offenders cause they're bringing the average down. 

The best solution is probably in the tests for those two.

shellscape:
👍 I'm going to just add a bunch of coverage fixes to this PR. generally like to keep that separate but since we're failing in fine fashion here it can't hurt.

lib/linters/qualifying_element.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
             }
 
             result.endsWith = selector.last.type;
-
+            console.log(selector.nodes);
```
wareczek: @shellscape common ;)

shellscape: nooooo I can't believe I let that slip in. we need to lint for that. good catch

shellscape: @wareczek https://github.com/lesshint/lesshint/pull/214

PR #219, opened 2016-07-20
Added brackets around .lesshintrc sample in README

JoshuaKGoldberg:
It's a little confusing that the first example of a `.lesshintrc` file is not actually valid.


shellscape:
nice subtle catch there!

PR #220, opened 2016-07-23
fixes #216. enable linting of root node (used by finalNewline and trailingWhitespace)

### LINKED ISSUE #216: finalNewline and trailingWhitespace linter bugs
Since moving over to PostCSS it doesn't look as though these two linters have been firing.

Their lint functions don't seem to be triggered when using `nodeTypes: ['root']`. Changing this to `nodeTypes: ['rule']`, does fix this but the internal logic for these looks to be incorrect atm.

garethwhittaker:


lib/linter.js
```diff
@@ -192,6 +170,34 @@ var runChecks = function (ast, fullPath, lines, config) {
...
+        var key = Object.keys(option)[0];
+
+        option[key] = merge(option[key], {
+            line: node.source.start.line
+        });
+
+        lineOptions.push(option);
+    });
+
+    checkNode(ast.root);
```
jwilsson: Could you add a comment here, explaining why it's necessary (just for future reference).

jwilsson:
One tiny comment. Other than that, LGTM.

garethwhittaker:
updated :smiley:

jwilsson:
Great! Just waiting to see if anyone else has any comments.

shellscape:
I'm good with it.

jwilsson:
Released in `2.0.1` (ping @JoshuaKGoldberg for NuGet package).

JoshuaKGoldberg:
@jwilsson uploaded!

Issue #222, opened 2016-07-27
spaceAroundOperator doesn't like prefixed functions


tkazec:
```css
test {
    background-image: -webkit-linear-gradient(xxx);
}
```

```
Warning: test.less: line 2, col 19, spaceAroundOperator: Operators should be preceded and followed by one space.
```

shellscape:
this is a bug with the postcss-values-parser. looking into it.

shellscape:
I've added another test to the spaceAroundOperators specs just to assert this case, but the commit listed above to postcss-values-parser fixes this issue. Re-install lesshint or force the latest version of postcss-values-parser in node_modules and you should be set.

tkazec:
Awesome, thanks a bunch!!

PR #223, opened 2016-07-27
adding spec for #222 assertion

shellscape:


shellscape:
@jwilsson appveyor is such a punk. why's it always failing?

jwilsson:
@shellscape I don't know, I'll restart it and see if that helps. Not really satisfied with it but haven't found a replacement :/

shellscape:
@jwilsson it doesn't seem to be picking up the new version of postcss-values-parser

jwilsson:
It seems it won't invalidate the `node_modules` cache, I'll try to invalidate it with something like this: https://www.appveyor.com/docs/build-cache#cache-dependencies

jwilsson:
That did the trick 🎉 

Merging!

PR #229, opened 2016-07-30
Add blankLineAfterBlock linter

### LINKED ISSUE #116: Create blankLineAfterDeclaration linter
We'd like to be able to enforce a blank new line after a declaration. For example:
```less
// bad
.thing1 {
    color: red;
}
.thing2 {
    color: blue;
}
```

```less
// good
.thing1 {
    color: red;
}

.thing2 {
    color: blue;
}
```

jwilsson:
I named it `blankLineAfterBlock` since it's actually checking blocks and not declarations (petty thing, I know).

Closes #116.


lib/linters/blank_line_after_block.js
```diff
@@ -0,0 +1,32 @@
...
+            /**
+             * Hopefully node.ruleWithoutBody will be implemented on AtRule nodes
+             * in the future: https://github.com/webschik/postcss-less/issues/55
+             */
+            return;
+        }
+
+        /**
+         * If it's the first node, ignore it since we actually check
+         * if the node is preceded by en an empty line
```
garethwhittaker: By checking the node is preceded by an empty line, we are losing the option of having comments above a block unless we additionally check and cater for this:

```css
.foo {
    width: 100%;
}

// comment
.bar {
    height: 100%;
} 
```

jwilsson: I knew I forgot about something! I'll see if it's possible to solve it without an extra check (but I doubt it, though to how PostCSS gives us the nodes).

shellscape:
I humbly suggest `newlineAfterBlock` as an alternative name for this linter.

Should we allow options for the number of 'blank lines' or newlines?

Code looks solid to me.

jwilsson:
Found (and fixed) a bug with nested blocks and comments.

I'm fine with either name, lets see what some other people think.

I think we can hold off with a option for the number of empty lines for now at least, or would a `no_line` option be enough?

garethwhittaker:
👍 for `newlineAfterBlock`

jwilsson:
I'm guessing we settled on `newlineAfterBlock` and no other options for now? If so, I'll change the name and merge it.

PR #233, opened 2016-08-01
Added code examples of how to set "Custom configuration path" and "Ho…

KennyLindahl:
…w to use the default reporter"

Basically what i wanted to do is adding simple code examples of what i think most people would need.
There was no documentation that there was a promise returned from .checkFile() etc. and that should be reflected in a simple example i believe.

Later on: Maybe we would like to add the possibility to pass a path to .configure() and not only just the config object, so we don't need to do the fs.readFileSync() stuff i show in the first code example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@
 * [Requirements](#requirements)
 * [Installation](#installation)
 * [Configuration](#configuration)
-* [Code examples](#code-examples)
 * [Custom linters](#custom-linters)
 * [CLI usage](#cli-usage)
 * [Reporters](#reporters)
@@ -44,6 +43,20 @@ Each option is then specified by its own JSON object, for example:
         "style": "one_space" // Comments are allowed
     }
 }
+```
+
+### Setting custom configuration path
+
+```js
+var path = require('path');
+var fs = require('fs');
+var LessHint = require('lesshint');
+var lesshint = new LessHint();
+
+var lesshintConfigPath = path.resolve('my-custom-path/.lesshintrc');
+var lesshintConfig = JSON.parse(fs.readFileSync(lesshintConfigurePath, 'utf8'));
+
+lesshint.configure(lesshintConfiguration);
 ```
 
 ### Inline configuration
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,6 +118,34 @@ It's also possible to define your own linters to add to the built-in list. These
...
+### Setting custom configuration path
+
+```js
+var path = require('path');
+var fs = require('fs');
+var LessHint = require('lesshint');
+var lesshint = new LessHint();
+
+var lesshintConfigPath = path.resolve('my-custom-path/.lesshintrc');
+let lesshintConfig = JSON.parse(fs.readFileSync(lesshintConfigurePath, 'utf8'));
```
shellscape: mixing var and let?

README.md
```diff
@@ -117,6 +118,34 @@ It's also possible to define your own linters to add to the built-in list. These
...
+
+lesshint.configure(lesshintConfiguration);
```
shellscape: `lesshintConfig` instead, right?

README.md
```diff
@@ -117,6 +118,34 @@ It's also possible to define your own linters to add to the built-in list. These
...
+```
+
+### Using the default reporter
```
jwilsson: Could we merge this with "Using reporters from code" further down?

Cause I noticed know that that example is wrong so we might as well replace it with this one and a comment explaining that the reporter name is optional.

README.md
KennyLindahl: Yep, will change it.

README.md
KennyLindahl: Thanks, i use ES6 with babel in my project and it should not be in the example.

README.md
KennyLindahl: Yes, but then i will remove the "Code examples" title and move "Setting custom configuration path" under the "Configuration" section. Makes sense right?

jwilsson: Yeah, sounds good!

KennyLindahl:
Please have a look again @jwilsson :)

jwilsson:
LGTM, thanks for the contribution!

PR #235, opened 2016-08-02
Update gulp-mocha to 3.x

jwilsson:
The title says it all. Nothing more, nothing less.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "gulp-coveralls": "^0.1.4",
     "gulp-eslint": "^2.0.0",
     "gulp-istanbul": "^0.10.2",
-    "gulp-mocha": "^2.1.3",
+    "gulp-mocha": "^3.0.0",
     "rewire": "^2.3.3",
     "rimraf": "^2.4.2",
     "sinon": "^1.14.1"


KennyLindahl:
A recommendation is to always specify specific versions to ensure that it's always stable.
Eventually the libs will be incompatible or change API.

Maybe we want to do this change in this PR as well?

Also it looks like the AppVeyor build failed, or is this expected in certain Node versions?
Would be good to get rid of it in any case.

shellscape:
@KennyLindahl have to disagree with the use of caret - that's the preferred modifier to use these days. https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004 "Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. ... Caret ranges are ideal when an author may make breaking changes between 0.2.4 and 0.3.0 releases"

KennyLindahl:
@shellscape It's a recommendation, so it can be ignored if you don't agree. However i speak from my own experience, i've seen it break multiple times because of this.

shellscape:
if someone is ignoring semver, then yes it can break. that's pretty edge-casey to go worrying about when it's so widely used, the default that NPM uses out of the box, and the likelihood that someone at mocha will ignore semver best practices (they'd be lambasted by the community).

KennyLindahl:
Yes, it is edge case, but it is safer. But go ahead and ignore if you don't agree.

jwilsson:
@KennyLindahl Thanks for the input. I don't know why AppVeyor failed (it just hung and didn't even try to install anything), a restart and everything is fine.

But I think it's OK to run with `^` which we do for everything else too. If something were to break (which I highly doubt) we'll deal with it then. But again, thanks for the input, highly appreciated!

I'll merge this then since there doesn't seem to be any other comments.

PR #237, opened 2016-08-08
fixes #236: import options breaking importPath linter

### LINKED ISSUE #236: Rule `importPath` with `filenameExtension = true` complains about inline import
If I have the rule `importPath` with `"filenameExtension": true`, and use an inline import, lesshint complains that the import should include the file extension even when it does.

E.g. this import:

```
@import (inline) 'variables.less';
```

Gives this error:

```
Warning: main.less: line 13, col 9, importPath: Imported file, "(" should include the file extension.
```

I'm using the currently latest version of lesshint, 2.0.2.

shellscape:
This is just a simple regex replace on the `params` node property. It should hold us over until https://github.com/webschik/postcss-less/issues/57 is fixed. Not an ideal fix by any means, but the likelihood of the import options list growing faster than we can update this regex is slim to none.

Added a few dupe tests with an import option just to make sure it doesn't break moving forward.

### DIFF
--- a/lib/linters/import_path.js
+++ b/lib/linters/import_path.js
@@ -15,15 +15,21 @@ module.exports = {
 
     lint: function importPathLinter (config, node) {
         var results = [];
+        var importOptsRx = /(\s+)?\((\s+)?(reference|inline|less|css|once|multiple|optional)(\s+)?\)(\s+)?/gi;
         var value;
         var file;
         var ast;
+        var params;
 
         if (node.name !== 'import') {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


jwilsson:
Not happy that we need it but the code LGTM. I'll see if I can put some work into `postcss-less`

shellscape:
@jwilsson shall we merge now for a quick fix or wait for postcss-less?

jwilsson:
Yeah, I'll merge it now and we can just remove it later.

PR #238, opened 2016-08-08
fixes #225: multiple urls in urlFormt, urlQuotes

### LINKED ISSUE #225: urlFormat and urlQuotes only checks the first URL in a declaration
When declaring multiple URLs, for example:
```less
background-image: url(http://example.com/img/image.jpg), url(http://example.com/img/image2.jpg);
```

With this config:
```json
{
    "urlFormat": {
        "enabled": true,
        "style": "relative"
    },
    "urlQuotes": {
        "enabled": true
    }
}
```

Only the first URL will be reported by both linters.

shellscape:


lib/linters/url_format.js
```diff
@@ -19,30 +19,45 @@ module.exports = {
             }
         };
         var uri = ast.first.first;
```
jwilsson: Couldn't we move this down into the traversing? And remove the assignment.

lib/linters/url_quotes.js
```diff
@@ -8,45 +8,59 @@ module.exports = {
...
     lint: function urlQuotesLinter (config, node) {
+        var results = [];
         var rSingle = /^\'(.+)\'$/;
         var rDouble = /^\"(.+)\"$/;
+        var self = this;
         var column;
         var value;
         var ast;
         var uri;
+        var func;
```
jwilsson: Same thing as above, move it into the iteration perhaps? Just to contain it a little bit better.

jwilsson:
Some nits, otherwise LGTM.

jwilsson:
Awesome 😃

PR #248, opened 2016-09-09
Space before brace new line and whitespace

### LINKED ISSUE #244: SpaceBeforeBrace New Line
I think there is a bug in the reg ex here: https://github.com/lesshint/lesshint/blob/77a502cb25dd8e8368b9a35aaabd183352d6c03e/lib/linters/space_before_brace.js#L15

It does not allow for whitespace between the selector and the brace so this is allowed

```css
ul
{
//Some Styles
li
{
//Some Styles
}
}
```

But this is not because between li and { there are 4 spaces.
```css
ul
{
    //Some Styles

    li
    {
        //Some Styles
    }
}
```

I think the fix would be to change the regex to /^\n *\t*$/ which would require a return directly after the selector but also allow for tabs or a spaces used to indent the code.

I do not think this is intended behaviour and I am happy to send a pull request but but fixing it would be a breaking change.

crash-dive:
This fixes #244 where setting "space before brace" to "new_line" did not allow you to indent the code.

It also fixes another issue in the "one_space" reg ex which used \s instead of space literal. \s does not mean space it means all whitespace characters so it would also pick up new lines (providing you use Unix style of just LF and not windows which uses two characters) and tabs.

Test have been provided and all pass.

### DIFF
--- a/lib/linters/space_before_brace.js
+++ b/lib/linters/space_before_brace.js
@@ -12,9 +12,9 @@ module.exports = {
     lint: function spaceBeforeBraceLinter (config, node) {
         var column;
         var styles = {
-            'new_line': /^\n$/,
+            'new_line': /^\n *\t*$/,
             'no_space': /^$/,
-            'one_space': /^\s$/
+            'one_space': /^ $/
         };
 
         if (config.style && !styles[config.style]) {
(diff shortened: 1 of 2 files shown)


shellscape:
👍

jwilsson:
Thanks for contributing!

PR #253, opened 2016-09-12
Ignore errors from bad symlinks

### LINKED ISSUE #251: Fails when encounters a dead symlink
This happens a lot in my meteor app. Meteor creates a symlink inside the `.meteor` directory, then later becomes dead, and then lesshint won't work unless I manually remove it (which is not possible with our CI system).

``` bash
mkdir -p /tmp/lesshint/styles/
cd /tmp/lesshint/
npm i lesshint
touch styles/style.less
ln -s $(pwd)/deadlink styles/deadlink
```

```
$ ls -l styles/
lrwxr-xr-x  1 user  group  21 Sep 12 11:31 deadlink -> /tmp/lesshint/deadlink
```

```
$ ./node_modules/.bin/lesshint styles/
An unkown error occured, please file an issue with this info:
Error: ENOENT: no such file or directory, stat 'styles/deadlink'
    at Error (native)
```

runarberg:
Fixes #251

### DIFF
--- a/lib/lesshint.js
+++ b/lib/lesshint.js
@@ -17,22 +17,29 @@ Lesshint.prototype.checkDirectory = function (checkPath) {
         files = files.map(function (file) {
             var fullPath = path.join(checkPath, file);
 
-            return fs.stat(fullPath).then(function (stats) {
-                if (stats.isDirectory()) {
-                    return this.checkDirectory(fullPath);
-                }
-
-                // Check if the file should be excluded
-                if (this.isExcluded(fullPath)) {
+            return fs.isSymLink(fullPath).then(function (isSymLink) {
+                if (isSymLink) {
+                    // Ignore symlinks
                     return [];
                 }
 
-                // Check if the file has the correct extension
-                if (!this.hasAllowedExtension(file)) {
-                    return [];
-                }
+                return fs.stat(fullPath).then(function (stats) {
+                    if (stats.isDirectory()) {
+                        return this.checkDirectory(fullPath);
+                    }
 
-                return this.checkFile(fullPath);
+                    // Check if the file should be excluded
+                    if (this.isExcluded(fullPath)) {
+                        return [];
+                    }
+
+                    // Check if the file has the correct extension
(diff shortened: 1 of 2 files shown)


runarberg:
I'm sorry but your test coverage came a bit down after this. Not sure how to test that errors are re-thrown.

runarberg:
This could have been easily fixed by using [`fs.lstat()`][fs.lstat] instead of [`fs.stat()`][fs.stat]. However, the [vow-fs] doesn't port that method.

[fs.lstat]: https://nodejs.org/api/fs.html#fs_fs_lstat_path_callback
[fs.stat]: https://nodejs.org/api/fs.html#fs_fs_stat_path_callback
[vow-fs]: https://github.com/dfilatov/vow-fs

shellscape:
👍

jwilsson:
Thanks for working on this!

I'm not sure if just blindly ignoring all symlinks is the way to go. I see three possible solutions.

1. I did some quick testing and while a `vow-fs.exists()` method exists, it doesn't follow symlinks but just checks whether the thing passed to it exists. However, the core `fs.access()` method does follow symlinks and will allow us to check if regular files exist too. Unfortunately, `vow-fs` doesn't expose `access` but we can always check if that's something that could be added to that library. If not, it's easy to add that check ourselves using the core `fs` module.

2. We add a option via `.lesshintrc` called `ignoreSymlinks` (or similar of course) which defaults to `false` for backward compatibility reasons. Meaning it won't ignore symlinks by default.

3. We simply ignore all symlinks.

shellscape:
hm good point. worth contributing the lstat method port to vow?

runarberg:
I think the second option is the most sane one. Perhaps some users will be pleased to discover that some of their symlinks are dead, while other users actually want to lint a file that a symlink points to (can't think of a reason why, but one must exist).

runarberg:
Created a new PR #255 for option 2 from @jwilsson last comment.

jwilsson:
Closing this since `excludedFiles` will work as expected in `3.0`.

PR #255, opened 2016-09-14
Add `ignoreSymlinks` config

### LINKED ISSUE #251: Fails when encounters a dead symlink

runarberg:
Fixes #251

### DIFF
--- a/lib/config/defaults.json
+++ b/lib/config/defaults.json
@@ -3,6 +3,8 @@
 
     "excludedFiles": [],
 
+    "ignoreSymlinks": false,
+
     "attributeQuotes": {
         "enabled": true,
         "style": "single"
--- a/lib/lesshint.js
+++ b/lib/lesshint.js
@@ -17,22 +17,28 @@ Lesshint.prototype.checkDirectory = function (checkPath) {
         files = files.map(function (file) {
             var fullPath = path.join(checkPath, file);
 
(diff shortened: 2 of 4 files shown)


jwilsson:
Sorry about the late reply, been out of town.

But AppVeyor doesn't seem to like it, I don't know if it's due to something with symlinks on Windows?

runarberg:
Sorry for the inactivity. Just rebased from upstream.

My guess for reason of the appveyor failure is that there is something wrong with how postcss handles symlinks on windows.

jwilsson:
Closing this since `excludedFiles` will work as expected in `3.0`.

Issue #257, opened 2016-09-15
Less merge (comma or space) should be excluded from "duplicateProperty"


MichelGratton:
In LESS there is feature use to merge property with a comma or a space by adding respectively "+" or "+_". These properties can be duplicated inside a single bloc. In my case it's inside a mixins. I can add all properties using this feature in the exclude property. But I think, this should be automaticaly exclude for all properties ending with "+" or "+_".

See: http://lesscss.org/features/#merge-feature

shellscape:
Please provide some sample LESS

MichelGratton:
Here two very basic mixins to test it. First example will merge two box-shadow property with a comma and the second mixin will merge to property with a space.
```less
.sample-merge-comma(@addInset: false) {
    box-shadow+: 0 0 10px #555;
    & when not (@addInset = false) {
        box-shadow+: inset 0 0 5px #222;
    }
}
```
---
```less
.sample-merge-space(@addRotate: false) {
    transform+_: scale(2);
    & when not (@addInset = false) {
        transform+_: rotate(15deg);
    }
}
```

PR #258, opened 2016-09-16
Add support for LESS merge feature

MichelGratton:
DuplicateProperty config should ingnore duplicate property ending with a "+" or "+_". This is a LESS merge feature. See: http://lesscss.org/features/#merge-feature

### DIFF
--- a/lib/linters/duplicate_property.js
+++ b/lib/linters/duplicate_property.js
@@ -19,7 +19,7 @@ module.exports = {
                 return true;
             }
 
-            if (properties.indexOf(decl.prop) >= 0) {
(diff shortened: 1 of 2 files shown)


shellscape:
The code change looks good, but we'll need some specs to go along with this.

MichelGratton:
This PR fix this Issue: https://github.com/lesshint/lesshint/issues/257 I think all informations is there

shellscape:
I think you're misunderstanding. When we make a new code change, there needs to be specs (tests) created for the change. New specs for the property merging should be created in this file: https://github.com/lesshint/lesshint/blob/master/test/specs/linters/duplicate_property.js

PR #261, opened 2016-09-30
Report invalid inline options

wohanley:
As of 2.1.1 an invalid inline option, (say, `// lesshint fnord`) will be reported unhelpfully by lesshint as `TypeError: Cannot read property 'line' of undefined`. This fixes the Error construction in parseRules to report the problem more clearly.

Incidentally, it seems to my amateur eye that it wouldn't be terribly difficult to report this inline with other errors, the same way CSS parse errors are handled. I could take a crack at that too, if it seems like a useful feature.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -57,7 +57,7 @@ var parseRules = function (node) {
     try {
         rules = JSON.parse('{' + rules + '}');
     } catch (e) {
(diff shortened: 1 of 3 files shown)


jwilsson:
LGTM, thanks for contributing!

Could you file a separate issue to discuss whether we should report these errors together with regular ones? Thanks!

PR #264, opened 2016-10-10
Account for trailing whitespace in newlineAfterBlock

### LINKED ISSUE #245: newlineAfterBlock with nested blocks
Hello,

if got problems to resolve the following warning(s)

```
Warning: app.less: line 1636, col 5, newlineAfterBlock: All blocks should be followed by a new line.
Warning: app.less: line 1641, col 3, newlineAfterBlock: All blocks should be followed by a new line.
Warning: app.less: line 1644, col 5, newlineAfterBlock: All blocks should be followed by a new line.
Warning: app.less: line 1649, col 3, newlineAfterBlock: All blocks should be followed by a new line.
Warning: app.less: line 1652, col 5, newlineAfterBlock: All blocks should be followed by a new line.
Warning: app.less: line 1657, col 3, newlineAfterBlock: All blocks should be followed by a new line.
Warning: app.less: line 1660, col 5, newlineAfterBlock: All blocks should be followed by a new line.
```
It works not on nested blocks for me. Here is my less source: (starting from line 1631)

```
/* Animations */
[ui-view] {
  &.ng-enter {
    opacity: 0;
    transition: 0.25s linear all;
    &.ng-enter-active {
      opacity: 1;
    }
  }

  &.ng-add {
    opacity: 0;
    transition: 0.25s linear all;
    &.ng-add-active {
      opacity: 1;
    }
  }

  &.ng-leave {
    opacity: 1;
    transition: 0.25s linear all;
    &.ng-leave-active {
      opacity: 0;
    }
  }

  &.ng-remove {
    opacity: 1;
    transition: 0.25s linear all;
    &.ng-remove-active {
      opacity: 0;
    }
  }
}
```

I found nothing about nested formatting in your documentation, so I tried some cases but always got

`newlineAfterBlock: All blocks should be followed by a new line.`

or

`trailingWhitespace: There should't be any trailing whitespace.`

How should the less code structured to remove the warning? Or is it a bug?

jwilsson:
Allow trailing whitespace to be present on the empty line. This will hopefully fix #245.

### DIFF
--- a/lib/linters/newline_after_block.js
+++ b/lib/linters/newline_after_block.js
@@ -6,6 +6,7 @@ module.exports = {
     message: 'All blocks should be followed by a new line.',
 
     lint: function newlineAfterBlockLinter (config, node) {
+        var regexp = /\n[ \t]*\n?/;
         var parent = node.parent;
         var prev;
 
@@ -35,12 +36,12 @@ module.exports = {
          * unless it's the first comment of a block.
          */
         if (prev.type === 'comment') {
-            if (Object.is(prev, parent.first) || prev.raws.before.indexOf('\n\n') !== -1) {
+            if (Object.is(prev, parent.first) || !regexp.test(prev.raws.before)) {
                 return;
             }
         }
 
-        if (node.raws.before.indexOf('\n\n') === -1) {
+        if (!regexp.test(node.raws.before)) {
             return [{
                 message: this.message
             }];
--- a/lib/linters/newline_after_block.js
+++ b/lib/linters/newline_after_block.js
@@ -36,7 +36,7 @@ module.exports = {
          * unless it's the first comment of a block.
          */
         if (prev.type === 'comment') {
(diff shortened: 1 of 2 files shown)


tkazec:
Personally, I think the way you're building source strings is fragile and will likely cause further issues in the future, because it's not representative of real world use. Can you use multiline strings that preserve formatting? (Plenty of ways to do so – but my favorite is `[].join('\n')`.)

shellscape:
six of one, half dozen of another regarding string building - religious debate relegated to style guides imho.

👍  from me

tkazec:
Technically the way the string is built does not matter, but at least the result should reflect expectations. The original test case did not catch this issue because the formatting in the code did not match the actual content.

jwilsson:
I don't think any test would have caught this, I wouldn't have thought about a line containing trailing whitespace and accounting for that anyway. But I can agree that building strings like this isn't optimal, but we can solve it with template literals once we drop support for older Node versions.

BTW, PRs for enhancing things are always welcome! 😉

PR #273, opened 2016-10-24
Added a shorthand `true` for `{ enabled: true }`

### LINKED ISSUE #272: Allow .lesshintrc options to be true instead of { "enabled": true }
This works:

```json
	"attributeQuotes": {
		"enabled": true
	},
```

This does no:

```json
	"attributeQuotes": true,
```

Would it be as simple as changing the `if (!options || (options && !options.enabled)) {` under the `ast.root.walk` under `linter.js`' `runChecks`?

JoshuaKGoldberg:
Fixes #272.

Rules can just say they're `true` rather than the full `{ enabled: true
}`.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -107,7 +107,7 @@ var runChecks = function (ast, fullPath, lines, config) {
             var lint;
 
             // Bail if the linter isn't wanted
-            if (!options || (options && !options.enabled)) {
+            if (!options || (options && options !== true && !options.enabled)) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


jwilsson:
Thanks!

@JoshuaKGoldberg I'd like to invite you to the `lesshint` org if there's no objections. Ping @lesshint/owners.

JoshuaKGoldberg:
No objections here.

kokarn:
More people means I can do even less!!! 🍰

Issue #149, opened 2016-03-02
Max Warnings Configuration


shellscape:
Do we have a max warnings configuration I'm not seeing? Wiring this up to gulp (through gulp-lesshint) I'm trying to force gulp to exit with 1 if there's any warnings at all. Certainly something I could add to the gulp plugin, but was thinking that lesshint itself should return 1 through the cli and something falsey if used programmatically. Thoughts?

jwilsson:
Absolutely, sounds great!

jwilsson:
@shellscape Do you wanna do it or should we mark it as "Help wanted"?

shellscape:
was just thinking about this today. I can take care of it, but probably won't be until next week sometime. busy weekend ahead!

PR #277, opened 2016-11-03
Add a max-warnings cli flag

runarberg:
Where number below `max-warnings` result in a zero exit code. This is
desirable for projects that allow a certain amount of warnings to pass
the build system, and the build system that relies on a zero exit code
for the build to pass.

We default to 0, for backwards compatibility. A max warning of -1 will
allow any number of warnings to pass.

### DIFF
--- a/bin/lesshint
+++ b/bin/lesshint
@@ -19,6 +19,7 @@ program
     .option('-e, --exclude [pattern]', 'path pattern to exclude')
     .option('-l --linters [linters]', 'paths to custom linters')
     .option('-r, --reporter [reporter]', 'reporter to use (<PATH>|stylish), defaults to "stylish"')
+    .option('-x, --max-warnings [int]', 'Number of warnings to trigger nonzero exit code, defaults to 0')
     .parse(process.argv);
 
 cli(program);
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -61,7 +61,9 @@ module.exports = function (program) {
     promises = program.args.map(lesshint.checkPath, lesshint);
     Vow.all(promises).then(function (results) {
         var hasError;
+        var maxWarnings;
         var reporter;
+        var warnings;
 
         results = [].concat.apply([], results);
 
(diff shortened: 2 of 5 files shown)


jwilsson:
@shellscape What's your take on this? And related to #149?

bin/lesshint
```diff
@@ -19,6 +19,7 @@ program
     .option('-e, --exclude [pattern]', 'path pattern to exclude')
     .option('-l --linters [linters]', 'paths to custom linters')
     .option('-r, --reporter [reporter]', 'reporter to use (<PATH>|stylish), defaults to "stylish"')
+    .option('--max-warnings [int]', 'Number of warnings to trigger nonzero exit code, defaults to 0')
```
shellscape: let's follow the same argument pattern here. I'd suggest `-x`

lib/cli.js
```diff
@@ -86,7 +88,23 @@ module.exports = function (program) {
         if (hasError) {
             exitDefer.reject(EXIT_CODES.ERROR);
         } else {
-            exitDefer.reject(EXIT_CODES.WARNING);
+            maxWarnings = parseInt(program.maxWarnings, 10) || 0;
+
+            if (maxWarnings === -1) {
+                exitDefer.resolve(EXIT_CODES.OK);
+
```
shellscape: nitpick: needed line break?

test/specs/cli.js
```diff
@@ -77,6 +77,54 @@ describe('cli', function () {
         });
     });
 
+    it('should exit with a non-zero status code when lint warnings pass `--max-warnings`', function () {
```
shellscape: I'd like to see one more test where the max warnings is set to 2, and the warnings generated are 3. or some such. just for the sake of being thorough.

shellscape:
@jwilsson yeah that'd close that issue alright

jwilsson:
Thanks @runarberg! 🎉

Issue #283, opened 2016-11-14
Split the CLI into a new class


JoshuaKGoldberg:
As per https://github.com/lesshint/lesshint/issues/281, it would be useful to be able to call the CLI's logic via `require`/`import` rather than some form of subprocess.

Required for #278.

jwilsson:
Makes sense, I guess. 

I'd be happy to help out/discuss design etc.

JoshuaKGoldberg:
@jwilsson it looks like most of `cli.js` could just be moved into its own class in a new file.

Thinking of a name is hard. `runner.js`?

PR #284, opened 2016-11-14
Allow reporters to be objects, not just file paths

### LINKED ISSUE #282: Allow reporters to be objects, not just file paths
When running `lesshint` via `requires`/`imports` in code, instead of the CLI, it would be useful to pass a reporter object instead of a file path to one. This would allow for stateful reporters in tools that require `lesshint`.

The actual change would be pretty small, too - in `Lesshint.prototype.getReporter`, something like: 

```javascript
    // Object with .report as a method: return it directly
    if (reporter.constructor !== 'string' && 'report' in reporter) {
        return reporter;
    }
```

Required for https://github.com/lesshint/lesshint/issues/278.

JoshuaKGoldberg:
Fixes #282.

I held off from documenting in this PR since until #283 is resolved I don't believe it would be easy for users to use this.

### DIFF
--- a/lib/lesshint.js
+++ b/lib/lesshint.js
@@ -127,6 +127,11 @@ Lesshint.prototype.getReporter = function (reporter) {
         return require(reporterPath);
     }
 
+    // Object with .report as a method: return it directly
+    if (reporter.constructor !== 'string' && reporter.report) {
+        return reporter;
+    }
+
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
Will investigate build breaks.

jwilsson:
LGTM

PR #286, opened 2016-11-14
Added a JSON reporter

### LINKED ISSUE #280: Come with a JSON reporter?
Seems like a pretty useful reporter for external tools; is there much problem in adding one by default?

```javascript
/*eslint no-console: 0*/

'use strict';

module.exports = {
    name: 'json',
    report: function report (results) {
        results.forEach(function (result) {
            console.log(JSON.stringify(result));
        });
    }
};
```

JoshuaKGoldberg:
Logs stringified results as JSON.

Fixes #280.

### DIFF
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -1,0 +1,12 @@
+/*eslint no-console: 0*/
+
+'use strict';
+
+module.exports = {
+    name: 'json',
+    report: function report (results) {
+        results.forEach(function (result) {
+            console.log(JSON.stringify(result));
+        });
+    }
+};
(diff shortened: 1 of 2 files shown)


jwilsson:
Just a small thing I thought of now, do we really want to print a line per object and not just an array with everything?

JoshuaKGoldberg:
Instead of:

```javascript
results.forEach(function (result) {
    console.log(JSON.stringify(result));
});
```

switching to:

```javascript
console.log(JSON.stringify(results));
```

Seems simpler to me. :+1:

jwilsson:
Let's go for it!

JoshuaKGoldberg:
Went for it.

PR #288, opened 2016-11-15
Absolute positions

### LINKED ISSUE #287: Report absolute line position
For tools (like automutate) that manipulate files as `string[]`s rather than `string[][]` it can be useful to have the absolute position of a complaint, in addition to line & column number.

I think we just compute these by creating a new `linesRaw` that has lines *with* their trailing newline characters.

Required for #278.

JoshuaKGoldberg:
Fixes #287.

It's computed by keeping the "raw" (trailing endlines included) lines from inputs.

Note that these are 0-based indices instead of 1-based like columns and lines.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -91,7 +91,7 @@ var getInlineOptions = function (node) {
     return parseRules(node);
 };
 
-var runChecks = function (ast, fullPath, lines, config) {
+var runChecks = function (ast, fullPath, lines, linesRaw, config) {
     var filename = path.basename(fullPath);
     var lineOptions = [];
     var results = [];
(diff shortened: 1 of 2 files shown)


jwilsson:
Hmm, I'm just thinking whether it's better to be consistent with the indices or not. 

On one hand, the `position` should only be read by the computer (and I kind of think that it should be 0-based) but I'd like a second opinion on it. Ping @shellscape @kokarn.

JoshuaKGoldberg:
A note in favor: this is necessary for accurate autolinting and probably necessary for other tools that would want to hook into this type of info.

kokarn:
I think column & line should be 1-based. 
It just makes sense "line 1" and "column 1" compared to the alternatives. 
But I can understand the reasoning for going with 0-base as well as long as they are consistent.

Issue #281, opened 2016-11-14
Allow an onExit other than the exit package


JoshuaKGoldberg:
When trying to run `lesshint` via...

```javascript
const lesshint = require('lesshint/lib/cli');
const settings = { /* ... */ };

lesshint(settings)
    .then(statusCode => console.log(`Finished with status ${statusCode}`})
    .catch(error => console.error(`Error: ${error}`));
```

...nothing gets logged in the happy case because [exit](https://www.npmjs.com/package/exit) is called at the end.

If we change one line in cli.js, it becomes possible to provide a custom `onExit` callback:

```javascript
    exitPromise.always(function (status) {
        // exit(status.valueOf());
        (program.onExit || exit)(status.valueOf());
    });
```

it would be possible to provide a custom onExit.

```javascript
const settings = {
    onExit: () => {}
};
```

Required for https://github.com/lesshint/lesshint/issues/278.

In case you're wondering why I'm not using the recommended `require('lesshint')`: it's to get all the features of the CLI without having to spawn subprocesses. A more complicated alternative would be to refactor out the non-`exit`-y parts of the CLI from `cli.js`.

kokarn:
👍

shellscape:
hm. when I take a look at other tools which have both a programmatic and cli interface, I can't come up with another one that supports calling the cli programmatically. gulp and grunt are two that come to mind. I'd be more inclined to work the essential bits into the main code from the cli, rather than open up and/or encourage programmatic use of the cli file. that just seems like an anti-pattern to me.

JoshuaKGoldberg:
@shellscape agreed. I was being lazy. :)

https://github.com/palantir/tslint/pull/1688 is a similar construct.

JoshuaKGoldberg:
Filed #283.

JoshuaKGoldberg:
Fix would be in `lib/lesshint.js`:

```diff
     // Nothing defined, just fall back to our simple default
-    if (!reporter) {
+    if (!reporter || reporter === 'default') {
```

PR #289, opened 2016-11-15
Moved most of cli.js into a new runner.js

### LINKED ISSUE #283: Split the CLI into a new class

JoshuaKGoldberg:
Fixes #283.

### DIFF
--- a/lib/lesshint.js
+++ b/lib/lesshint.js
@@ -121,7 +121,7 @@ Lesshint.prototype.getReporter = function (reporter) {
     var reporterPath;
 
     // Nothing defined, just fall back to our simple default
-    if (!reporter) {
+    if (!reporter || reporter === 'default') {
         reporterPath = path.resolve(__dirname, './reporters/default');
 
         return require(reporterPath);
--- a/lib/lesshint.js
+++ b/lib/lesshint.js
@@ -13,6 +13,12 @@ var Lesshint = function () {
 };
 
 Lesshint.prototype.checkDirectory = function (checkPath) {
+    if (this.isExcluded(checkPath)) {
+        return new Vow.Promise(function (resolve) {
+            resolve([]);
+        });
+    }
+
     return fs.listDir(checkPath).then(function (files) {
         files = files.map(function (file) {
             var fullPath = path.join(checkPath, file);
@@ -49,6 +55,12 @@ Lesshint.prototype.checkFile = function (checkPath) {
 };
 
 Lesshint.prototype.checkPath = function (checkPath) {
+    if (this.isExcluded(checkPath)) {
+        return new Vow.Promise(function (resolve) {
+            resolve([]);
+        });
+    }
+
     return fs.stat(checkPath).then(function (stats) {
         if (stats.isDirectory()) {
             return this.checkDirectory(checkPath);
@@ -125,6 +137,11 @@ Lesshint.prototype.getReporter = function (reporter) {
         reporterPath = path.resolve(__dirname, './reporters/default');
 
         return require(reporterPath);
+    }
+
+    // Object with .report as a method: return it directly
+    if (reporter.constructor !== 'string' && reporter.report) {
+        return reporter;
     }
 
     // Try to find it somewhere on disk
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -1,5 +1,3 @@
-/*eslint no-console: 0*/
-
 'use strict';
 
 var configLoader = require('./config-loader');
@@ -27,9 +25,10 @@ Runner.prototype.run = function () {
     var config;
 
     if (!this.settings.args.length) {
-        console.error('No files to lint were passed. See lesshint -h');
-
-        exitDefer.reject(EXIT_CODES.NOINPUT);
+        exitDefer.reject({
+            code: EXIT_CODES.NOINPUT,
+            errors: ['No files to lint were passed. See lesshint -h']
+        });
 
         return exitPromise;
     }
@@ -48,9 +47,10 @@ Runner.prototype.run = function () {
             config.linters.push.apply(config.linters, this.settings.linters);
         }
     } catch (e) {
-        console.error("Something's wrong with the config file. Error: " + e.message);
-
-        exitDefer.reject(EXIT_CODES.CONFIG);
+        exitDefer.reject({
+            code: EXIT_CODES.CONFIG,
+            errors: ['Something\'s wrong with the config file. Error: ' + e.message]
+        });
 
         return exitPromise;
     }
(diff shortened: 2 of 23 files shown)


JoshuaKGoldberg:
@shellscape is this what you meant by moving the essential bits into the main code?
@jwilsson is this design something you're ok with? *(you mentioned being up for discussing it)*

lib/runner.js
```diff
@@ -0,0 +1,117 @@
...
+
+Runner.prototype.run = function () {
+    var lesshint = new Lesshint();
+    var exitDefer = Vow.defer();
+    var exitPromise = exitDefer.promise();
+    var promises = [];
+    var config;
+
+    if (!this.settings.args.length) {
+        console.error('No files to lint were passed. See lesshint -h');
```
jwilsson: Wouldn't it be better to reject the promise with a error message and a error code and keep all the `console.log`s in `cli.js`. So this file can be used from another Node program without any logging output?

lib/runner.js
```diff
@@ -0,0 +1,117 @@
...
+
+        if (this.settings.exclude) {
+            config.excludedFiles.push(this.settings.exclude);
+        }
+
+        if (this.settings.linters) {
+            config.linters.push.apply(config.linters, this.settings.linters);
+        }
+    } catch (e) {
+        console.error("Something's wrong with the config file. Error: " + e.message);
```
jwilsson: Same reasoning as above.

lib/runner.js
```diff
@@ -0,0 +1,117 @@
...
+        var warnings;
+
+        results = [].concat.apply([], results);
+
+        reporter = lesshint.getReporter(this.settings.reporter);
+
+        if (reporter) {
+            reporter.report(results);
+        } else {
+            console.error('Could not find reporter "%s".', this.settings.reporter);
```
jwilsson: Same reasoning as above.

lib/runner.js
```diff
@@ -0,0 +1,117 @@
...
+            }, 0);
+
+            if (warnings > maxWarnings) {
+                exitDefer.reject(EXIT_CODES.WARNING);
+            } else {
+                exitDefer.resolve(EXIT_CODES.OK);
+            }
+        }
+    }, this).fail(function (error) {
+        console.error('An unknown error occurred when checking "%s", please file an issue with this info:', error.lesshintFile);
```
jwilsson: Same reasoning as above.

jwilsson:
I have some thoughts about logging from inside the runner but other than that I'm good.

lib/runner.js
JoshuaKGoldberg: Agreed! Will change.

JoshuaKGoldberg:
The `'should exit without errors when passed a built-in reporter name'` spec test suddenly failing confused me for a while... but it looks like it's not actually an accurate test anymore?

> ```javascript
> result = cli({
>    args: [path.join(__dirname, '../data/files/ok.less')],
>    reporter: 'stylish'
> });
>
> return result.then(function (status) {
>     expect(status).to.equal(0);
> });
> ```

`'stylish'` isn't included with `lesshint`. The test asserts that the returned status is 0, but doesn't care that `console.error` was called during `lesshint` with `Could not find reporter "stylish".`. 

> ```javascript
> reporter = lesshint.getReporter(program.reporter);
>
> if (reporter) {
>     reporter.report(results);
> } else {
>     console.error('Could not find reporter "%s".', program.reporter);
> }
> ```

When `runner.js` started resolving and returning, the behavior changed to now correctly give an error code.

JoshuaKGoldberg:
Another change proposal for the CLI tests: `result.then(...)` instead of `result.fail(...)`.

jwilsson:
> Question from this: are you ok with no longer giving a 0 error status on an invalid linter name?
Yes, that seems like the correct behaviour.

> Another change proposal for the CLI tests: `result.then(...)` instead of `result.fail(...)`. 
Wouldn't it be better to separate the `then()` and `catch()` in `cli.js` instead? Cause' I'm guessing that's why, right? But still use `always()` for the exit code for example.

JoshuaKGoldberg:
Ok. Bad news, good news.

Bad news: it's been a month since you last commented. Sorry about that. PR updated & merge conflicts resolved.

Good news: I reverted the functional changes to the CLI such that it again resolves or rejects as appropriate. I *also* became very irritated with how the CLI failing to resolve or reject appropriately would cause errors with any subsequent test, so I made the runner and its constituent reporters take in a "`logger`" object (à la dependency injection). The gist of it is:
* `bin/lesshint` doesn't pass `exit` to `cli`; instead, it retrieves the result and calls `exit` with it.
* `lib/cli.js` takes in a `logger` instead of directly using `console` and passes that to its `Runner`.
* `lib/runner.js::Runner` takes in a `logger` and passes it to its reporters.
* reporters are now given a `logger` object after `errors` that they can (should) use to log.

Using a dependency injected `logger` gives the benefit that consumers of the reporters (such as our tests or external consumers of `lesshint`) don't need to globally mock or spy or stub the global console.

One downside is that existing custom linters won't be aware of the change, but since it's backwards compatible (doesn't change behavior for existing consumers) I don't think that should be a problem.

Thoughts?

shellscape:
> don't need to globally mock or spy or stub the global console.

In what situation would they ever have to mock/spy/stub `console.log`?

lib/lesshint.js
```diff
@@ -133,7 +133,7 @@ Lesshint.prototype.getReporter = function (reporter) {
     var reporterPath;
 
     // Nothing defined, just fall back to our simple default
-    if (!reporter) {
+    if (!reporter || reporter === 'default') {
```
shellscape: cherry pick this as a separate bug fix

JoshuaKGoldberg:
> In what situation would they ever have to mock/spy/stub console.log?

Generally any utility that would want to call the runner as a part of some other library. I'm thinking of [autolesshint](https://github.com/automutate/autolesshint) specifically.

```javascript
var cli = require('lesshint/lib/cli');

module.exports = function (logger) {
    return cli(
        {
            args: ['index.less'],
            config: '.lesshintrc'
        },
        logger);
};
```

> This is starting to look like a classic case of trying to do too much in a single PR.

Fair. The vision has expanded in scope from "split CLI into CLI and Runner" into "split CLI into CLI and *a bug-free when in standalone* Runner". Filing issues now...

jwilsson:
It looks like this PR has started to go off-track. Perhaps it's better to file a new one once we ironed out #281 and #300? Just my two-cents.

JoshuaKGoldberg:
:+1: will keep code around for easy viewing

PR #292, opened 2016-11-24
Restore console spies before assertions

### LINKED ISSUE #291: Restore console spies before assertions in tests
image

jwilsson:
Closes #291.

### DIFF
--- a/test/specs/cli.js
+++ b/test/specs/cli.js
@@ -38,9 +38,11 @@ describe('cli', function () {
         });
 
         return result.fail(function () {
-            expect(console.error.calledOnce).to.equal(true);
+            var called = console.error.calledOnce;
 
             console.error.restore();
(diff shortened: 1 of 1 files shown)


test/specs/cli.js
```diff
@@ -38,9 +38,11 @@ describe('cli', function () {
         });
 
         return result.fail(function () {
-            expect(console.error.calledOnce).to.equal(true);
+            var called = console.error.calledOnce;
```
jwilsson: Needed to move this around since `calledOnce` will be removed when restoring.

test/specs/cli.js
```diff
@@ -54,9 +56,11 @@ describe('cli', function () {
...
```
jwilsson: Same as above.

PR #303, opened 2016-12-20
Moved require('exit') to bin\lesshint

### LINKED ISSUE #281: Allow an onExit other than the exit package

JoshuaKGoldberg:
This allows consumers of the CLI function to not randomly close when the exit package is called.

Fixes #281.

### DIFF
--- a/bin/lesshint
+++ b/bin/lesshint
@@ -1,7 +1,6 @@
 #! /usr/bin/env node
 
 var cli = require('../lib/cli');
-var exit = require('exit');
 
 /**
  * Command line interface for lesshint.
@@ -24,5 +23,5 @@ program
     .parse(process.argv);
 
 cli(program).always(function (status) {
-    exit(status.valueOf());
+    process.exitCode = status.valueOf();
 });
--- a/bin/lesshint
+++ b/bin/lesshint
@@ -1,6 +1,7 @@
 #! /usr/bin/env node
 
 var cli = require('../lib/cli');
+var exit = require('exit');
 
 /**
  * Command line interface for lesshint.
@@ -22,4 +23,4 @@ program
     .option('-x, --max-warnings [int]', 'Number of warnings to trigger nonzero exit code, defaults to 0')
     .parse(process.argv);
 
-cli(program);
+cli(program, exit);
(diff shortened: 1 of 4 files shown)


JoshuaKGoldberg:
I'll push a second commit that, in addition to fixing the build breaks, changes the CLI to not interact with an onExit at all. Instead `bin/lesshint` willl do that logic.

lib/cli.js
```diff
@@ -16,15 +15,15 @@ var EXIT_CODES = {
     CONFIG: 78
 };
 
-module.exports = function (program) {
+module.exports = function (program, onExit) {
```
shellscape: ehhh thoughts on emitting an event instead @jwilsson @kokarn ?

bin/lesshint
```diff
@@ -22,4 +23,6 @@ program
     .option('-x, --max-warnings [int]', 'Number of warnings to trigger nonzero exit code, defaults to 0')
     .parse(process.argv);
 
-cli(program);
+cli(program).always(function (status) {
+    exit(status.valueOf());
```
jwilsson: I say we use `process.exitCode` instead. We only used `exit` to work around some strange output issue in an old Node version.

JoshuaKGoldberg:
Re the failing coverage: would you like me to add a `test/bin.js` that exercises `bin/lesshint`?

jwilsson:
@JoshuaKGoldberg Sure, that would be great! No rush :)

PR #307, opened 2016-12-23
Linter fix suggestions

### LINKED ISSUE #278: Auto-fix Lesshint bugs?
I’d like to make a project to automatically apply fixes for lesshint failures. A big blocker from adopting more linting for a lot of teams (mine included) is the time it takes to fix existing failures. I’ve started a  [rough draft](https://github.com/joshuakgoldberg/automutator) project that takes in and applies waves of suggested fixes, such as from a linter like lesshint.

A potential technical implementation could be:

1. Allow for an optional `suggestFix` method for linters that takes in failures and outputs some POJO representation of the fix
    * My proposal for the format of such as fix: [automutate](https://github.com/JoshuaKGoldberg/automutate) ([abstract mutation spec](https://github.com/JoshuaKGoldberg/automutate/blob/master/src/mutation.ts) / [sample mutator](https://github.com/JoshuaKGoldberg/automutate/blob/master/src/mutators/textDeleteMutator.ts))
2. Add a new reporter that just reports these fix suggestions as JSON

What are your thoughts on this? Is it something you’d be interested in having me work on?

*Edit: Discovered issues*:
* #287: Lesshint reports line and column number for errors, but thanks to file inconsistencies with `\n` verses `\r\n` endlines those aren't directly mappable to absolute character positions in files. This would require remembering file endlines and using them to output absolute `"position": number`s in error complaints.
* #283: In order to get the full CLI logic, you need to run `cli.js`. This needs to either take in a different `onExit` than the `exit` package or have its logic refactored out.
* ~~#284: Custom reporter objects should be allowed, so that we don't have to go with one that directly `console.log`s. Alternately passing a stream that reporters can stream to (with console logging as the default) could work.~~ Never mind - works!

JoshuaKGoldberg:
Linters can now provide a `suggestFix` method. It should take in their lint and return an `automutate`-style fix suggestion. These are added to lint pieces as `suggestedFix`.

Implemented for the following rules:
- [x] attribute_quotes
- [x] border_zero
- [x] comment
- [ ] decimal_zero
- [ ] depth_level
- [ ] duplicate_property
- [ ] empty_rule
- [x] final_newline
- [ ] hex_length
- [ ] hex_notation
- [ ] hex_validation
- [ ] id_selector
- [ ] import_path
- [ ] important_rule
- [ ] max_char_per_line
- [ ] newline_after_block
- [ ] property_ordering
- [ ] property_units
- [ ] qualifying_element
- [ ] selector_naming
- [ ] single_line_per_property
- [ ] single_line_per_selector
- [ ] space_after_property_colon
- [ ] space_after_property_name
- [ ] space_after_property_value
- [ ] space_around_comma
- [ ] space_around_operator
- [ ] space_before_brace
- [ ] space_between_parens
- [ ] string_quotes
- [x] trailing_semicolon
- [x] trailing_whitespace
- [ ] url_format
- [x] url_quotes
- [ ] zero_unit

Fixes #278.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -172,7 +172,7 @@ var runChecks = function (ast, fullPath, lines, linesRaw, config) {
                         + linesRaw.slice(0, piece.line - 1).join('').length
                         - 1;
 
-                    if (config.suggestFixes && linter.suggestFix) {
+                    if (linter.suggestFix) {
                         piece.suggestedFix = linter.suggestFix(piece, options, node);
                     }
 
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -173,7 +173,7 @@ var runChecks = function (ast, fullPath, lines, linesRaw, config) {
                         - 1;
 
                     if (linter.suggestFix) {
-                        piece.suggestedFix = linter.suggestFix(piece, options, node);
+                        piece.suggestedFix = linter.suggestFix(piece, options, node, linesRaw);
                     }
 
                     results.push(piece);
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -172,6 +172,10 @@ var runChecks = function (ast, fullPath, lines, linesRaw, config) {
                         + linesRaw.slice(0, piece.line - 1).join('').length
                         - 1;
 
+                    if (config.suggestFixes && linter.suggestFix) {
+                        piece.suggestedFix = linter.suggestFix(piece, options, node);
+                    }
+
                     results.push(piece);
                 });
             }
@@ -248,7 +252,7 @@ exports.lint = function (input, fullPath, config) {
     lines = input.split('\n');
 
     try {
-        parser = this.getParser(input);
+        parser = exports.getParser(input);
         ast = parser.root.toResult();
 
         results = runChecks(ast, fullPath, lines, linesRaw, config);
--- a/lib/linters/attribute_quotes.js
+++ b/lib/linters/attribute_quotes.js
@@ -35,5 +35,33 @@ module.exports = {
         if (results.length) {
             return results;
         }
+    },
+
+    suggestFix: function fixAttributeQuotesFailure (lint) {
+        var end = lint.position + (lint.source.indexOf(']') - lint.source.indexOf('='));
+
+        return {
+            mutations: [
+                {
+                    insertion: '\'',
+                    range: {
+                        begin: lint.position
+                    },
+                    type: 'text-insert'
+                },
+                {
+                    insertion: '\'',
+                    range: {
+                        begin: end
+                    },
+                    type: 'text-insert'
+                }
+            ],
+            range: {
+                begin: lint.position,
+                end: end
+            },
+            type: 'multiple'
+        };
     }
 };
--- a/lib/linters/border_zero.js
+++ b/lib/linters/border_zero.js
@@ -51,12 +51,19 @@ module.exports = {
     },
 
     suggestFix: function fixBorderZeroFailure (lint, config) {
-        var length = config.style === 'zero'
-            ? 'none'.length
-            : '0'.length;
+        var insertion;
+        var length;
+
+        if (config.style === 'zero') {
+            insertion = '0';
+            length = 'none'.length;
+        } else {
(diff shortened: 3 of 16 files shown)


JoshuaKGoldberg:
As of 85fed651c20a409c065f6b65ad98b05dbbfd3222 (commit 2) there are two things that could use improvement:
* `linter.suggestFix` is always called if it exists. This might not be good if advanced rules end up having computationally complex `suggestFix`. I'd wanted to have a `.suggestFixes` param in the program but couldn't figure out how to pass it down to `linter.js`' `runChecks` without a bunch of changes.
* The added `util.js` test method, `suggestFixes`, is a lot more complex than the existing `parse`. Linters need the full line position for their lint pieces, so it needs to call the logic that's currently in `linter.js`' `runChecks`. I couldn't find a (better/faster) way than blindly calling `lint` without major refactors.

Are any of you maintainers able to assist?

shellscape:
May be an outlier here: I'm not a fan of adding this to each individual linter. I can see the value of callbacks/events in strategic places to allow for *external* extensions/add-ons/plugins to provide this kind of functionality, but I don't feel the proper home is in a linter library. I'd much rather see lesshint become extensible, rather than the size and complexity of the library increase.

jwilsson:
@shellscape Do you have an idea/example on how to solve these sort of things with callbacks/events?

JoshuaKGoldberg:
One problem with separating complaints and fixes is their logic getting out of sync. When a rule's configuration settings are expanded, for example, the maintainer+code for suggesting fixes then has to be notified+updated. Complex changes in configuration settings could even introduce version conflicts which the suggestions code would then have to deal with.

If code bloat is also a concern, a strategy for minimizing bloat could be to use helpers that create the (admittedly verbose) POJO format. TSLint does something similar: [example](https://github.com/palantir/tslint/blob/e63d0db49b1a42f5518434f5acf683be1489b91b/src/rules/trailingCommaRule.ts#L247).

On the other hand, if you end up deciding against adding suggestions to the core library, autolesshint would be fine (just more complex and prone to paragraph 1). The pseudocode logic would switch from:

```javascript
while (true) {
    const suggestions = lesshint(settings).map(complaint => complaint.suggestion);
    if (!suggestions.length) break;

    for (const suggestion of orderAndRemoveOverlaps(suggestions)) {
        fixSuggestion(suggestion);
    }
}
```

...to:

```javascript
while (true) {
    const complaints = lesshint(settings).map(complaint);
    if (!suggestions.length) break;

    const suggestions = complaints.map(complaint => createSuggestion(complaint.name, complaint);
    for (const suggestion of orderAndRemoveOverlaps(suggestions)) {
        fixSuggestion(suggestion);
    }
}
```

shellscape:
@jwilsson I'd personally like to see lesshint emit an event for each "complaint" (error/warning) either before, during, or after reporting. the differences in code that @JoshuaKGoldberg has outlined there are negligible imho, so much so that I would rather opt for not adding to lesshint to accommodate a difference of two lines. I know I'm minimizing (but not snarkily so) the impact and differences but I just don't see the benefit of adding another piece of complexity to the lib when it can be easily remedied outside.

jwilsson:
I just wanna start off by saying I'm still very torn on this proposal and I still like the idea of autofixing things.

On one hand, I would want to see this kind of functionality in `lesshint` but using the AST we already have from PostCSS, modifying it, and stringifying the results (I don't know how much of this is possible, though).

But adding this amount of code to `lesshint` in order to support an external tool does seem a bit much I'm afraid. We'll of course promote it in the `README` and it does indeed not look like it's too much extra work in `autolesshint` (absolutely not trying to minimize anything in any way).

I hope to have some time to help out with `autolesshint` in any way I can and I'm not against the earlier changes we've made to `lesshint` to accommodate both this and possibly other tools. If we can find a nice way to expose events and/or callbacks I would be interested in that as well.

@JoshuaKGoldberg Thanks for being so patient and working with us!

JoshuaKGoldberg:
No worries! I'll close this PR and cut + paste the suggestions to work on it separately. The really important one is #320.

PR #311, opened 2016-12-30
ES6 rewrite

jwilsson:
It is huge, so the quick breakdown is:

* Changed all `var`s to `let` and `const`. Moved most variable declarations around too to take advantage of block scoping etc.
* Removed all use of `var self = this` and replaced it with arrow functions.
* The main `Lesshint` class is now a "proper" class.
* Some new ESLint rules, I'm open for discussion on them of course 😉.
* No test changes other than `let`/`const` so this should be completely backwards compatible.
* I also managed to shorten `maxCharPerLine` a bit.

I'd like to refactor `linter.js` in the future as well.


shellscape:
for consideration, the .eslintrc I use on almost all of my projects:

```json
{
  "env": {
    "es6": true,
    "mocha": true,
    "node": true
  },
  "rules": {
    "brace-style": [2, "stroustrup", { "allowSingleLine": true }],
    "consistent-return": [0],
    "complexity": [0],
    "indent": [2, 2, { "SwitchCase": 1, "VariableDeclarator": 1 }],
    "max-len": [1, 1000],
    "no-bitwise": [0],
    "no-extra-parens": [0],
    "one-var": [2, {
      "var": "always",
      "let": "always",
      "const": "never"
    }],
    "semi": [2, "always"],
    "space-before-function-paren": [0],
    "space-unary-ops": [0]
  },
  "parserOptions": {
    "ecmaVersion": 8
  }
}
```

I'd also like to recommend that the rule values be converted to declarations like `"indent": ["error", 2]` as they're MUCH easier to read. I've been lazy about doing that to my own configs.

shellscape:
And for good measure, some slightly stricter rules I use for koa-webpack, which uses babel:

```json
{
  "parser": "babel-eslint",
  "rules": {
    "consistent-return": [0],
    "complexity": [0],
    "space-before-function-paren": [0],
    "indent": [2, 2, { "SwitchCase": 1, "VariableDeclarator": 1 }],
    "max-len": [1, 1000],
    "brace-style": [2, "stroustrup", { "allowSingleLine": true }],
    "space-unary-ops": [0],
    "one-var": ["error", "always"],
    "no-extra-parens": [0],
    "no-bitwise": [0],
    "semi": [2, "always"],
    "prefer-arrow-callback": ["error", { "allowNamedFunctions": true }]
  },
  "env": {
    "mocha": true,
    "node": true
  }
}
```

PR #313, opened 2017-01-10
Replace Vow with native Promises

jwilsson:
N/A

I'd appreciate an extra set of eyes on `lib/lesshint.js` since that's where the biggest changes are.


shellscape:
🎉

PR #316, opened 2017-01-11
fix(package): update postcss-values-parser to version 1.0.2

jwilsson:
Closes #315

https://greenkeeper.io/

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "postcss": "^5.0.19",
     "postcss-less": "^0.14.0",
     "postcss-selector-parser": "^2.0.0",
-    "postcss-values-parser": "^0.1.7",
+    "postcss-values-parser": "^1.0.2",
     "rcfinder": "^0.1.8",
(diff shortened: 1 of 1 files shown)


shellscape:
@jwilsson those `TokzenError`s are a result of changes meant to follow the css values spec. Looks like most are to do with operators, for which there are pretty strict validating guidelines. I can loosen that up with an `options` object or some such for Lesshint though.

jwilsson:
@shellscape Yeah, I think some sort of "loose" mode is needed. It's failing on things like `10px+ 10px` which is valid in Less.

shellscape:
cool. so `options.loose` that won't throw the errors and allow parsing through could probably work. I'll try and make that happen today. if you get a sec and want to throw a PR at it, I'll dance a jig and chug a beer.

jwilsson:
Closing this pending completion of https://github.com/lesshint/postcss-values-parser/issues/15.

PR #320, opened 2017-01-18
Move CLI functionality to its own Runner class

jwilsson:
This is basically a refresh of #289, after all the other `Promise`-changes and ES6 rewrites.

N/A

I'm contemplating whether or not this should be exposed in the public API, which would introduce a breaking change.

Right now, we only expose `lib/lesshint.js` in the `main` field in `package.json`.

What I'm thinking is whether we should do like this instead:

_package.json_
```js
{
...
"main": "lib/index.js",
...
}
```

_lib/index.js_
```js
module.exports = {
  Lesshint: require('./lesshint'),
  Runner: require('./runner')
};
```

So to use it, instead of
```js
const Lesshint = require('lesshint');
```

One would use
```js
const Lesshint = require('lesshint').Lesshint;
```

Thoughts on that? @kokarn @shellscape @JoshuaKGoldberg


lib/runner.js
```diff
@@ -0,0 +1,119 @@
...
+                });
+            }
+
+            Promise.all(promises).then((results) => {
+                results = [].concat.apply([], results);
+
+                reporter.report(results);
+
+                if (!results.length) {
+                    return resolve(EXIT_CODES.OK);
```
JoshuaKGoldberg: Do you want to return `{ status: EXIT_CODES.ERROR }`?

JoshuaKGoldberg:
Re exposing the runner in the public API - is there a reason against `require("lesshint/lib/runner");`?

lib/runner.js
```diff
@@ -0,0 +1,119 @@
...
+                    return reject({
+                        message: '',
+                        status: EXIT_CODES.ERROR
+                    });
+                }
+
+                const maxWarnings = parseInt(this.options.maxWarnings, 10) || 0;
+
+                if (maxWarnings === -1) {
+                    return resolve(EXIT_CODES.OK);
```
JoshuaKGoldberg: Do you want to `return resolve({ status: EXIT_CODES.OK });`?

lib/runner.js
jwilsson: Yes, we should do that for consistency. Thanks!

lib/runner.js
jwilsson: Yes, we should do that for consistency. Thanks!

jwilsson:
IMO, using `require("lesshint/lib/runner");` feels a bit too "hacky", like you're loading something that shouldn't be loaded by an external actor. But if we document it properly I guess I could live with that 😁

JoshuaKGoldberg:
Haha yeah. I was hesitant to propose you change the project for my requirements... but after some thought I also prefer `.Lesshint`.

```javascript
import { Lesshint, Runner as LesshintRunner } from "lesshint";
```

```javascript
import * as lesshint from "lesshint";
const runner = new lesshint.Runner({...});
```

lib/runner.js
kokarn: Should we also include a message? I see we return an object with an empty message in other places

lib/runner.js
kokarn: Should we include a message here as well? Or do we separate `resolve()` and `reject()` in that regard?

lib/runner.js
```diff
@@ -0,0 +1,119 @@
...
+                }, 0);
+
+                if (warnings > maxWarnings) {
+                    return reject({
+                        message: '',
+                        status: EXIT_CODES.WARNING
+                    });
+                }
+
+                resolve(EXIT_CODES.OK);
```
kokarn: `return resolve({ status: EXIT_CODES.OK });` here as well I assume. Also see previous comments about message

jwilsson:
@kokarn I think we should follow the "Rule of Silence" on success, we won't have anything to say anyway.

PR #323, opened 2017-01-21
Silently ignore fs errors

jwilsson:
This is a follow-up to #313.
 
Earlier, we would just silently ignore any fs errors so I think it's best to keep this behaviour. And rejecting everything because of one inaccessible file seems a bit harsh too.

N/A

Not really.


shellscape:
Does this still provide the necessary output for users to determine that a failure occurred?

jwilsson:
@shellscape Hmm, I'm unsure what you mean. With this approach, inaccessible files will just be silently ignored.

I did, however, perform some more thorough testing of the current (`2.4.0` behaviour) and while we currently don't care about the file being accessible or not, newer Node versions will complain about the unhandled rejected `Promise`. 

Before this patch, Node would still complain. Perhaps it's better to handle the rejected `Promise` and rethrow an error about the inaccessible file.

kokarn:
I'd also like some sort of output or any way to determine that a file is inaccessible.

My reasoning is that it's probably because of a misconfiguration and the user probably want's to know that.

shellscape:
Perhaps pushing those file paths that are bad to an array and a small section in the output after the results listing the files that couldn't be processed?

kokarn:
I like that. Keeping it minimal but not hiding what could be important information.

jwilsson:
I checked how Stylelin` and ESLint handles it, and they both just exit with an error as soon as a inaccessible file is encountered. My vote's for that, just exit with an error showing but showing the name of the inaccessible file.

jwilsson:
The last commit is still kinda a work in progress. Some tests are needed too, but just to let you guys see where I'm going.

jwilsson:
I've added some tests (still a small decrease in coverage but this is about as much I can do without getting into really funky tests). 

I think this approach is fine, we can always add a list of inaccessible files at a later time.

PR #324, opened 2017-01-22
Documentation update

kloots:
I was looking to use custom linters in my project and found one small error in the sample custom linter code in the `README` so I fix it. Additionally, it wasn't clear from the `README` that custom linters defined in `lesshintrc` must also be enabled. My impression was they are enabled by default as a result of their definition, so I updated the docs to call they must also be enabled.


jwilsson:
Thanks for contributing! 🎉

PR #326, opened 2017-01-24
Expose Lesshint and Runner classes in the public API

jwilsson:
N/A

Nothing in particular.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,0 +1,6 @@
+'use strict';
+
+module.exports = {
+    Lesshint: require('./lesshint'),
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
Random thought: would it be at all beneficial to use a lazy-instantiated getter for `index.js`?

```javascript
'use strict';

module.exports = {
    Lesshint: require('./lesshint'),
    Runner: require('./runner')
};
```

I've never seen another project do this, but it could save on `require` times for someone using only one...

Otherwise LGTM 👍

Issue #331, opened 2017-01-28
New VariableOnly linter


justtal:
I have created a new linter called "variableOnly" that validate that values of specific property is a variable.

At our team for color and font-size (for now) we use only variables that defined in app variables file.

Do you think it can be useful for other? do you want me to commit it to lesshint or create my own custom linter package?

You can pass to the linter the properties you want to be variable only:

```
"variableOnly": {
    "enabled": true,
    "properties": ['color', 'font-size']
}
```

Cheers,

Tal.

jwilsson:
Sounds good to me!

Lets hear some other thoughts too, though.

justtal:
Do you want me guys to commit it?

shellscape:
I'd like to suggest a rename to `valueVariable`. Should consider future improvements to this and combinations of restraints of the property value using variables:

- always a variable
- never a variable
- (default would be disabled and unopinionated)

jwilsson:
I'm fine with either name. A PR would be very much welcome!

justtal:
@shellscape I really can't see a situation that someone will want the "never" option.

WDYT?

PR #333, opened 2017-02-01
Variable Value linter

justtal:
New variable value linter.

#331

### DIFF
--- a/lib/linters/variable_value.js
+++ b/lib/linters/variable_value.js
@@ -3,11 +3,11 @@
 const util = require('util');
 
 module.exports = {
-    name: 'variableOnly',
+    name: 'variableValue',
     nodeTypes: ['decl'],
     message: 'The value of [%s] must be a variable.',
 
-    lint: function variableOnlyLinter (config, node) {
+    lint: function variableValueLinter (config, node) {
         if (!config.enabled) {
             return;
         }
--- a/lib/linters/variable_value.js
+++ b/lib/linters/variable_value.js
@@ -5,22 +5,33 @@ const util = require('util');
 module.exports = {
     name: 'variableValue',
     nodeTypes: ['decl'],
-    message: 'The value of "%s" must be a variable.',
+    message: '%s is not allowed in "%s" property.',
 
     lint: function variableValueLinter (config, node) {
-        if (config.properties.indexOf(node.prop) === -1) {
-            return;
+        const reg = /@/i;
+
+        if (config.never && config.never.indexOf(node.prop) !== -1) {
+            if (reg.test(node.value)) {
+                return [{
+                    column: node.source.start.column +
+                            node.prop.length +
+                            node.raws.between.length,
+                    line: node.source.start.line,
+                    message: util.format(this.message, 'Variable', node.prop)
+                }];
+            }
         }
 
-        // It's a variable, so it's allowed
-        if (/^@/i.test(node.value)) {
-            return;
+        if (config.always && config.always.indexOf(node.prop) !== -1) {
+            if (!reg.test(node.value)) {
+                return [{
+                    column: node.source.start.column +
+                            node.prop.length +
+                            node.raws.between.length,
+                    line: node.source.start.line,
+                    message: util.format(this.message, 'Non variable', node.prop)
+                }];
+            }
         }
-
-        return [{
-            column: node.source.start.column + node.prop.length + node.raws.between.length,
-            line: node.source.start.line,
-            message: util.format(this.message, node.prop)
-        }];
     }
 };
--- a/lib/linters/variable_value.js
+++ b/lib/linters/variable_value.js
@@ -8,10 +8,10 @@ module.exports = {
     message: '%s is not allowed in "%s" property.',
 
     lint: function variableValueLinter (config, node) {
-        const reg = /@/i;
+        const variableRegex = /@/i;
 
         if (config.never && config.never.indexOf(node.prop) !== -1) {
-            if (reg.test(node.value)) {
+            if (variableRegex.test(node.value)) {
                 return [{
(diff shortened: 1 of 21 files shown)


shellscape:
man that linter name rubs me wrong 😆

justtal:
:)

I don't care changing it name at all.

I just didn't understand the need of "never a variable" option.

lib/linters/variable_value.js
```diff
@@ -0,0 +1,29 @@
...
+
+const util = require('util');
+
+module.exports = {
+    name: 'variableValue',
+    nodeTypes: ['decl'],
+    message: 'The value of [%s] must be a variable.',
+
+    lint: function variableValueLinter (config, node) {
+        if (!config.enabled) {
```
jwilsson: This check isn't necessary, it's checked beforehand in `linter.js`.

lib/linters/variable_value.js
```diff
@@ -0,0 +1,29 @@
...
+            return;
+        }
+
+        if (config.properties.indexOf(node.prop) === -1) {
+            return;
+        }
+
+        if (/^@/i.test(node.value)) {
```
jwilsson: Could you add a comment explaining this line, i.e.
`// It's a variable, so it's allowed`

test/specs/linters/variable_value.js
```diff
@@ -0,0 +1,77 @@
...
+    describe('#variableValue()', function () {
+        it('should have the proper node types', function () {
+            const source = 'color: #ABC;';
+
+            return spec.parse(source, function (ast) {
+                expect(spec.linter.nodeTypes).to.include(ast.root.first.type);
+            });
+        });
+
+        it('should ignore if rule disabled', function () {
```
jwilsson: Not needed as per above.

lib/linters/variable_value.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const util = require('util');
+
+module.exports = {
+    name: 'variableValue',
+    nodeTypes: ['decl'],
+    message: 'The value of [%s] must be a variable.',
```
jwilsson: We usually put properties etc. in messages in double quotes.

test/specs/linters/variable_value.js
```diff
@@ -0,0 +1,77 @@
...
+                const result = spec.linter.lint(options, ast.root.first);
+
+                expect(result).to.be.undefined;
+            });
+        });
+
+        it('should ignore if property not for lint', function () {
+            const source = 'color: #333;';
+            const options = {
+                enabled: true,
```
jwilsson: Could you remove all the `enabled: true` since they aren't necessary.

jwilsson:
Thanks for working on this!

I left a few comments, but overall it looks good.

I think that @shellscape wants (and please correct me if I'm wrong) to have a "never" option because we usually allow linters to be really flexible (even if it might seem a bit crazy). And as I said earlier, I'm fine on either name.

justtal:
@jwilsson  Thanks for the review!

So do you want me to change the config to be:

```
{
  "enabled": true,
  "properties": {
    "color": "always",
    "font-size": "never"
  }
}
```

jwilsson:
Yeah, that looks like it would work.

@shellscape What do you think? And do you have a better explanation for the `never` option?

shellscape:
yeah in the spirit of forward-thinking and flexibility, I could foresee people wanting to designate that certain properties never have a variable used for the value. I can't imagine why, but we have other linters that enable stranger things. 

I would suggest the following setup, to reduce the amount of config that would have to be written for projects that want to specify a litany of options for this linter:

```js
{
  "enabled": true,
  "always": [ "color", "background", "font-family", ... ],
  "never": [ "font-size" ]
}
```

justtal:
Done. Please review.

Will it be in the 3.0 release?

lib/linters/variable_value.js
```diff
@@ -0,0 +1,37 @@
...
+
+const util = require('util');
+
+module.exports = {
+    name: 'variableValue',
+    nodeTypes: ['decl'],
+    message: '%s is not allowed in "%s" property.',
+
+    lint: function variableValueLinter (config, node) {
+        const reg = /@/i;
```
jwilsson: Perhaps name this something like `reVariable` or `variableRegex`.

test/specs/linters/variable_value.js
```diff
@@ -0,0 +1,104 @@
...
+            }];
+
+            return spec.parse(source, function (ast) {
+                const result = spec.linter.lint(options, ast.root.first);
+
+                expect(result).to.deep.equal(expected);
+            });
+        });
+
+        it('should allow non variable if in never list', function () {
```
jwilsson: I think this one made it in twice (see line 60).

test/specs/linters/variable_value.js
```diff
@@ -0,0 +1,104 @@
...
+            };
+
+            return spec.parse(source, function (ast) {
+                const result = spec.linter.lint(options, ast.root.first);
+
+                expect(result).to.be.undefined;
+            });
+        });
+
+        it('should not allow non variable if in always list', function () {
```
jwilsson: Could you also add a test for the opposite, i.e. "should allow variable if in always list".

jwilsson:
Just a few more things. Thanks for working with us!

Unfortunately, it won't make it into `3.0.0`, I think it's big enough as it is. It'll definitely make `3.1.0` though, which I'm hoping we'll get out the door pretty soon afterwards.

justtal:
@jwilsson Thanks for the review. Some stupid mistakes :\

Fixed.

3.1.0 would be great.

jwilsson:
Thanks for contributing! 🎉 

I'm hoping to have a release out later today.

PR #334, opened 2017-02-03
Final 3.0 CLI touches

### LINKED ISSUE #321: Max-Warnings Config Issue
**Which version of `lesshint` are you using?** 2.4.0

**How are you running `lesshint`? CLI, Node.js API, Grunt/Gulp plugin?**
CLI for testing. package.json run command in Node.js app

**What's your `.lesshintrc` configuration?** Default

**If you're reporting a bug, please show us some code that's failing.**
`lesshint --max-warnings=-1 ./path_to/styles/

**What did you expect to happen?**
I was expecting my build to succeed regardless of how many warnings I had.

**What's the actual result?**
I get an exit code of 1 and I received the following error when I ran the previous code:
`error: unknown options '-1'`

While I could use a positive integer for the time being, it would be nice to know what I am doing incorrectly to receive this error or for it to be corrected to work as expected.

jwilsson:
* Reject `Runner` with real `Error`s.
* Fix handling of `--max-warnings=-1`. Closes #321.
* Fix a minor refactor typo with custom linters.

#321 among others.

@kokarn Is this what you had in mind? Hope I didn't step on any toes 😬

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const RunnerError = require('./errors/runner-error');
 const configLoader = require('./config-loader');
 const Lesshint = require('./lesshint');
 
@@ -20,10 +21,10 @@ class Runner {
     run () {
         return new Promise((resolve, reject) => {
             if (!this.options.args.length) {
-                return reject({
-                    message: 'No files to lint were passed. See lesshint -h',
-                    status: EXIT_CODES.NOINPUT
-                });
+                return reject(new RunnerError(
+                    'No files to lint were passed. See lesshint -h',
+                    EXIT_CODES.NOINPUT
+                ));
             }
 
             let config;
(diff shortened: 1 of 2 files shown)


lib/runner.js
```diff
@@ -98,10 +105,10 @@ class Runner {
                 }, 0);
 
                 if (warnings > maxWarnings) {
-                    return reject({
-                        message: '',
-                        status: EXIT_CODES.WARNING
-                    });
+                    return reject(new RunnerError(
+                        '',
```
kokarn: Can't we avoid this somehow?

jwilsson: Are you suggesting we put a message there? Or something else?

kokarn: Not sure what to do, just don't like passing empty messages I guess. 
Can't we have a default just to ease my mind or will that break something?

I guess it's stuff like ```if( error.message )``` that scares me a bit.

jwilsson: I don't wanna unnecessary output in case someone's piping the output somewhere else, reporters will already have printed everything that's needed.

kokarn: Hmm, I guess that makes sense. Just leaves me with a funny feeling.

Issue #266, opened 2016-10-11
Rule proposal: universalSelector


jwilsson:
Should work just like `idSelector`, disallow the use of universal selectors (i.e. `*`).

Should report:

``` less
* {
  color: red;
}

.foo * {
  color: red;
}
```

Shouldn't report:

``` less
a {
  color: red;
}

.foo a {
  color: red;
}
```

shellscape:
I like it.

kokarn:
Needs more cowbell/options? Do we always want to disallow this? Or just global scope?

deanc:
Hi @jwilsson - going to bring this one back from the dead too. It'd be nice to make this slightly more configurable also.

I would argue that such a rule in some cases might be nice. In fact it's usually the rule I write first in any CSS file to reset all browser styles to the same.

```less
* {
  box-sizing: border-box;
  margin:0;
  padding: 0;
}
```

Perhaps adding an option to allow the universal selector in the global scope, but disallow it in the form: `.foo *`.

jwilsson:
@deanc Could you open a new issue with that feature request, please?

Another option could be to add a `// lesshint-disable-next-line universalSelector` comment in those instances.

PR #336, opened 2017-02-04
Add universalSelector linter

jwilsson:
#266 

Nothing in particular.


lib/linters/README.md
```diff
@@ -835,6 +836,24 @@ Option     | Description
...
+## universalSelector
+Disallow the usage of the universal selector.
+The universal selector shouldn't be used since it'll cancel out any inheritance.
+
+### invalid
+```less
+* {
+    color: red;
+}
+```
```
justtal: consider add here another example with sub universal selector `.foo * {}` or change the * {} to a sub selector. It is not clear that also sub universal selectors are not allowed.

jwilsson: Good idea, I've added it.

PR #348, opened 2017-02-12
Mentioned autolesshint in the readme

### LINKED ISSUE #339: Mention autolesshint in the readme
3.0 is out, so I've resumed work on [autolesshint](https://github.com/automutate/autolesshint/). You can install it and try it out; it should work! Unfortunately its current rule coverage probably isn't enough to justify putting a mention on the lesshint readme. A few issue points here...
* At what point (if any) would you be willing to mention it?
* What features do you want in it?
* The list of rules that need either have a suggestions generator written or be documented as un-suggestible is at its [Complete rule coverage](https://github.com/automutate/autolesshint/milestone/1) milestone. I'd greatly appreciate help filling those out!

JoshuaKGoldberg:
Phrasing seems... fine? Very much open to changing it.

Fixes #339.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
 * [CLI usage](#cli-usage)
 * [Linters](#linters)
 * [Reporters](#reporters)
+* [`autolesshint`](#autolesshint)
 * [Developer resources](#developer-resources)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


jwilsson:
I was thinking that we add a list of complementary tools, Grunt/Gulp plugins, autolesshint, Atom plugin etc. How does that sound?

JoshuaKGoldberg:
SGTM! If you have a way you want to put it it might make sense for you to send a separate PR for everything. I'm not in a hurry.

jwilsson:
I added a list like this: https://github.com/lesshint/lesshint#complementary-tools instead.

PR #351, opened 2017-02-21
Replace istanbul with nyc

jwilsson:
Since `gulp-mocha 4.0` spawns a child process we need a coverage tool that handles that.
To get it working on older npm versions which doesn't dedupe by default, we need to explicitly include mocha as a dependency as well.

Thanks to @kokarn for the rubber ducking session!

N/A

Nope, it should be self-explanatory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.nyc_output
 coverage
 node_modules
 npm-debug.log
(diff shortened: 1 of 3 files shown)


shellscape:
LGTM. FWIW we don't _need_ to use gulp-mocha v4. It was something that solved a lot of problems using mocha programmatically. With the shift in tooling that node modules is seeing right now (where npm scripts are allllll the rage), it made sense. And with that in mind it probably makes sense to make this change. We're at least flexible whichever way the trend goes.

PR #353, opened 2017-02-28
Fixed linesRaw regexp to allow for duplicate `\n`s

### LINKED ISSUE #310: Double \n characters give incorrect finalNewline positions
**Which version of `lesshint` are you using?**
2.4.0

**How are you running `lesshint`? CLI, Node.js API, Grunt/Gulp plugin?**
Either CLI or Node.js API.

**What's your `.lesshintrc` configuration?**
`>node bin/lesshint --reporter lib/reporters/json temp/test.less`

**If you're reporting a bug, please show us some code that's failing.**
Contents of `temp/test.less`:
```less
.foo {}

.bar {}
```
To be precise, the string is `.foo {}\n\n.bar{}`

**What did you expect to happen?**
The JSON blob should report the position as 16 .

**What's the actual result?**
The position is reported as 22.

Note that the position gets correctly reported if a single `\n` is used (position 15) or `\r\n\r\n` is used (position 18).

**Would you be interested in submitting a PR for this issue?**
Yes.

JoshuaKGoldberg:
Regular expressions are hard!

Fixes #310.

### DIFF
--- a/lib/linter.js
+++ b/lib/linter.js
@@ -243,7 +243,7 @@ exports.resultSeverity = {
 };
 
 exports.lint = function (input, fullPath, config) {
-    const linesRaw = input.match(/[^\n]+(?:\r?\n|$)/g); // Include trailing line endings
+    const linesRaw = input.match(/([^\r\n]*(?:\r?\n)|[^\r\n]+)/g); // Include trailing line endings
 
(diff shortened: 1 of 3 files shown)


shellscape:
humbly requesting a test specifically for the multiple `\n`

Issue #357, opened 2017-03-02
Improve CLI usage


kokarn:
This is a placeholder for things I'd like to see implemented for the CLI.

* ```-v``` option to output a helping text that says "maybe you meant ```-V``` or ```--version```"
* Message when no files are found
* Something when just running ```lesshint```
* Allow glob patterns for files and/or folders

shellscape:
`globby` for the last one. best glob module I've used

jwilsson:
> -v option to output current version

We got `-V` and `--version` which is automatically supplied by `commander.js`. I also think we should save `-v` for a future `--verbose`.

> Message when no files are found

Do you mean when `lesshint foo/` is run and `foo/` is empty, it should print something? If so, agreed.

> Something when just running `lesshint`

This has worked before, I think we broke it in `3.0`.

> Allow glob patterns for files and/or folders

Agreed. It looks like we can replace `minimatch` (which we're currently using) with `globby` since `globby` wraps `glob` which uses `minimatch`. I'll do some further investigations.

jwilsson:
My bad, `globby` and `minimatch` does different things (of course). I say we use `globby` too.

kokarn:
Updated suggestion, until we get ```-v``` a helping text that says "maybe you ment ```-V```" would be nice

jwilsson:
I'm going to close this.

Point 3 and 4 are fixed and released.
Point 1 can't be solved in a nice way using `commander.js` and without adding a new flag.
Point 2 can't be really be solved either because there's really no way for us to know if there weren't any files found at all or just no linting errors in the files found.

PR #360, opened 2017-03-11
Handle both strings and error objects in custom errors

jwilsson:
Point 3 of #357, kinda. Since we broke that functionality in `3.0` with our custom errors which this commit fixes.

Not really.


jwilsson:
Didn't realize the coverage had gone down, added some tests.

PR #361, opened 2017-03-13
Update runner.js

Venomknight257:
Modified maxWarnings. new Boolean constructor is always returning true within a ternary despite having a false value. Functionality doesn't change, user still does not declare the --max-warnings flag to allow any number of warnings to pass through the build. If the --max-warnings flag has a value defined then it will break when limit is met.

Which issue, if any, does this resolve?
#321
Is there anything in this PR that needs extra explaining or should something specific be focused on?
Should be rather self explanatory.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -88,10 +88,13 @@ class Runner {
 
                 let maxWarnings;
 
+                //maxWarnings is either defined OR it is not.
+                //IF it is defined then set maxWarnings to the desired amount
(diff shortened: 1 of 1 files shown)


kokarn:
Could you have a look that the failing test? Seems like a node 4 issue.

jwilsson:
Closing this since the functionality here has changed. 

Thanks for working on it though, @Venomknight257!

Issue #364, opened 2017-03-15
urlQuotes throwing warning for LESS variable


amandakalk:
**Which version of `lesshint` are you using?**
3.1.0

**How are you running `lesshint`? CLI, Node.js API, Grunt/Gulp plugin?**
Gulp

**What's your `.lesshintrc` configuration?**

```js
{
    "fileExtensions": [".less"],
    "spaceAfterPropertyColon": {
        "enabled": true,
        "style": "one_space" // Comments are allowed
    },
    "propertyOrdering": {
        "enabled": false
    },
    "maxCharPerLine": {
        "enabled": false
    }
}
```

**If you're reporting a bug, please show us some code that's failing.**

```less
@brand-logo: '/assets/images/logo.svg';

.head-logo {
	background: url(@brand-logo);
}
```

**What did you expect to happen?**
No warning. The variable evaluates to a properly quoted url.

**What's the actual result?**
I get this warning: `urlQuotes: URLs should be enclosed in quotes`

Technically, I could write the background declaration as `background: url('@{brand-logo}');` to prevent the warning, but it feels weird to add unnecessary interpolation just to make the linter happy. Thoughts?

jwilsson:
Hi,
Thanks for the report!

I think it would make sense for it to ignore variables. If the rest of the team agrees, would you be interested in submitting a PR @amandakalk?

kokarn:
👍 

Would this come up somewhere else or maybe break in some situations? 
I'm looking at the [documentation](http://lesscss.org/features/#features-overview-feature-variables) and if I'm not mistaken you could end up doing something like

```less
@randomstring: aString;

#header {
  background-image: url( @randomstring );
}
```

Does this matter in any case or can we safely ignore all the vars?

amandakalk:
If you ignore variables completely, this would compile and not throw the error:
```less
@randomstring: test.png;

#header {
  background-image: url( @randomstring );
}
```
This **would not** compile at all (unquoted string with slash):
```less
@randomstring: assets/test.png;

#header {
  background-image: url( @randomstring );
}
```

So technically you could be ignoring a failure, but seems kind of unlikely that your URL wouldn't contain a slash in it.

jwilsson:
Thanks for the extra examples!

I agree that a URL without a slash seems unlikely, and we can never be sure of what the variable actually contains. I still think just ignoring variables in `url()` is fine (FWIW I checked other linters and they do it too).

[...]

PR #366, opened 2017-03-17
Add check for variable in url quotes linter.

amandakalk:
#364 

Seems pretty straightforward.

### DIFF
--- a/lib/linters/url_quotes.js
+++ b/lib/linters/url_quotes.js
@@ -34,6 +34,11 @@ module.exports = {
             const uri = child.first.next();
             const value = uri.value.trim();
 
+            // Ignore variables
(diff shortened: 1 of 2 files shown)


lib/linters/url_quotes.js
```diff
@@ -34,6 +34,11 @@ module.exports = {
             const uri = child.first.next();
             const value = uri.value.trim();
 
+            // Ignore variables
+            if (uri.value.substring(0, 1) === '@') {
```
jwilsson: Could you switch this to just `value.substring(...)`? `url( @bar )` will still error.

amandakalk: Sure, fixed it now.

jwilsson:
Cool, thanks for contributing! 🎉

Issue #372, opened 2017-04-12
Custom linters are not resolved when current directory is a child directory


WattyRev:
**Which version of `lesshint` are you using?**
3.1.1

**How are you running `lesshint`? CLI, Node.js API, Grunt/Gulp plugin?**
`lesshint *` from inside a child folder or via Atom lesshint plugin

**What's your `.lesshintrc` configuration?**

```js
{
    "fileExtensions": [".less", ".css"],

    "excludedFiles": [],

    // Custom linters
    "linters": [
        "lesshint-linters/color_variables.js" // NOTE, same result with ./lesshint-... or /leshint-...
    ],

    "colorVariables": {
        "enabled": true,
        "severity": "error"
    },

    // Learn about linters here: https://github.com/lesshint/lesshint/blob/master/lib/linters/README.md

    //... normal linter configs would be below here
}

```

**What did you expect to happen?**
I expected that my custom linter would get resolved when using the Atom lesshint plugin and when running `lesshint .` from a child folder.

**What's the actual result?**
When running `lesshint .`, an error is logged to the console:
```
An unknown error occurred when checking "/BitTitan/FrontEnd/bittitan-css/lib/bittitan-ui-kit/2.0/framework/framework.less", please file an issue with this info:
Error: Cannot find module 'lesshint-linters/color_variables.js'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.require.resolve (internal/module.js:16:19)
    at requireConfigLinter (/Users/swatson/.nvm/versions/node/v4.6.1/lib/node_modules/lesshint/lib/linter.js:234:32)
    at Array.map (native)
    at mergeLintersFromConfig (/Users/swatson/.nvm/versions/node/v4.6.1/lib/node_modules/lesshint/lib/linter.js:222:49)
    at runChecks (/Users/swatson/.nvm/versions/node/v4.6.1/lib/node_modules/lesshint/lib/linter.js:188:21)
    at Object.exports.lint (/Users/swatson/.nvm/versions/node/v4.6.1/lib/node_modules/lesshint/lib/linter.js:255:23)
    at Lesshint.checkString (/Users/swatson/.nvm/versions/node/v4.6.1/lib/node_modules/lesshint/lib/lesshint.js:107:29)
    at /Users/swatson/.nvm/versions/node/v4.6.1/lib/node_modules/lesshint/lib/lesshint.js:74:34
[...]

PR #373, opened 2017-04-12
If custom linter is not found, look for it based on location of config file | added colorVariables linter

WattyRev:

Resolves this issue: https://github.com/lesshint/lesshint/issues/372

The main idea here is that when we parse the config file, we store the path to that config file as _path, then we can use that when we look for custom linters to support custom linter paths that are relative to the .lesshintrc file.

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -14,7 +14,9 @@ const loadConfig = function (path) {
         data = data.slice(1);
     }
 
-    return JSON.parse(data);
+    data = JSON.parse(data);
+    data._path = path;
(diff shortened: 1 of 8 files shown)


kokarn:
I'm not a huge fan of having two separate things added in a single pull request.
I know it's a bit of work, but would you mind creating two separate pull requests? 
That way it's easier to look at and the history get's cleaned up a bit :)

WattyRev:
Separate pull requests added at #374  and #375.
Closing this PR in favor of those.

PR #374, opened 2017-04-13
Add color variables linter

WattyRev:
More related to #372 

This is adding a linter that prohibits hex colors unless used for variable assignment.

### DIFF
--- a/lib/linters/color_variables.js
+++ b/lib/linters/color_variables.js
@@ -26,6 +26,7 @@ module.exports = {
 
         const results = [];
         ast.first.walk((child) => {
+            // Only evaluate hex colors
             if (child.type !== 'word' || !/^#/.test(child.value)) {
                 return;
             }
@@ -35,12 +36,7 @@ module.exports = {
             if (/@[a-zA-Z0-9-_]*/.test(prop)) {
                 return;
             }
-
-            // Detect hex color
             const color = child.value;
-            if (/^#\d+$/.test(color)) {
-                return;
-            }
 
             // Hex color found, add to results
             results.push({
--- a/lib/linters/color_variables.js
+++ b/lib/linters/color_variables.js
@@ -18,12 +18,12 @@ module.exports = {
 
         const ast = parseValue(node.value, {
             loose: true
-        }).parse();
+        });
 
         const results = [];
         ast.first.walk((child) => {
             // Only evaluate hex colors
-            if (child.isColor) {
+            if (child.type !== 'word' || !child.isColor) {
                 return;
             }
 
--- a/lib/linters/color_variables.js
+++ b/lib/linters/color_variables.js
@@ -8,7 +8,7 @@ module.exports = {
     nodeTypes: ['decl'],
     message: '%s should be replaced with an existing variable.',
 
-    lint: function colorVariablesLinter(config, node) {
+    lint: function colorVariablesLinter (config, node) {
         /**
          * Just in case, since postcss will sometimes include the semicolon
          * in declaration values
@@ -21,11 +21,11 @@ module.exports = {
 
         // Parse the configuration for the style settings
         if (config.style) {
-            throw new Error(`Invalid setting value for colorVariables: ${config.style}`);
+            throw new Error(`Invalid setting value for colorVariables: ${ config.style }`);
         }
 
         const results = [];
-        ast.first.walk(child => {
+        ast.first.walk((child) => {
             if (child.type !== 'word' || !/^#/.test(child.value)) {
                 return;
             }
--- a/lib/linters/color_variables.js
+++ b/lib/linters/color_variables.js
@@ -1,0 +1,57 @@
+'use strict';
+
+const parser = require('postcss-values-parser');
+const util = require('util');
+
+module.exports = {
+    name: 'colorVariables',
+    nodeTypes: ['decl'],
+    message: '%s should be replaced with an existing variable.',
+
+    lint: function colorVariablesLinter(config, node) {
+        /**
+         * Just in case, since postcss will sometimes include the semicolon
+         * in declaration values
+         */
+        node.value = node.value.replace(/;$/, '');
+
+        const ast = parser(node.value, {
+            loose: true
+        }).parse();
+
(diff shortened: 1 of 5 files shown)


shellscape:
I think it's awesome that folks are creating their own linters. That said, I do still think this is too specialized to be in core. Rather, I think the code is sound and that this would make a great third-party linter module.

WattyRev:
What would be your suggestion on making it accessible as a third-party linter module? Is there an established method of doing that?

shellscape:
This https://github.com/Joshuakgoldberg/lesshint-browser-support is a good example (even though it's got TypeScript decorations) of a third party linter by @JoshuaKGoldberg.

I wouldn't close this PR just yet. This is only my opinion and the others should definitely weigh in first (and they may disagree with me)

jwilsson:
I like it and think it's a good fit for core since we already got `variableValue`.

What does the rest of @lesshint/team think?

_(I've only skimmed the code real quick so if we decide to go ahead there might be a few comments there, just a heads up)_

kokarn:
I'm not following. Doesn't the varibleValue rule do the same thing as this?

WattyRev:
The variableValue rule enforces the use of variables for certain properties, while this rule prohibits the use of hard-coded hex colors except when defining a variable. 

It is good for my team because we have a predefined color pallet, so we want to make sure people use the existing color variables rather than adding their own.

kokarn:
Ah, I think I see now. 

Instead of having variableValue configured for each property where you might add a colour you want to just make sure colours are always a var.

Just out of curiosity, how many different properties would it be for variableValue?

```less
background-color
border-color
color
fill
```

WattyRev:
Yes, but there are also the multi-part properies that would use color, like
background
border
box-shadow
etc. 

I think the variableValue rule is an all or none type thing, so mixing variables with other text doesnt fly

kokarn:
Hmm, I kinda like it and kinda don't. 
Maybe it's the limited scope that gets me.

Something that does this but also covers stuff like measurements (em, px, rem, %) and durations might be something that feels a bit more general.

But I'm very undecided I'm afraid. I'll let @jwilsson handle this.

PS: Sorry for the long back and forth :)

jwilsson:
@kokarn Could we get a Yes or No from you? 

Right now, we got one against and one in favor of, I'd want a majority for one or the other :)

kokarn:
I said I'd leave it up to you. If you say yes, then it's a yes from me :)

jwilsson:
Alright then, thanks @kokarn!

@WattyRev Really sorry for the delay on this one! Too much other things going on right now, but if you're still interested in finishing this up, I'll leave some small comments on the code. If not, we can still merge it and finish it up afterwards.

lib/linters/color_variables.js
```diff
@@ -0,0 +1,53 @@
...
+         * in declaration values
+         */
+        node.value = node.value.replace(/;$/, '');
+
+        const ast = parser(node.value, {
+            loose: true
+        }).parse();
+
+        // Parse the configuration for the style settings
+        if (config.style) {
```
jwilsson: We can remove this block, since we don't have any settings on this linter.

lib/linters/color_variables.js
```diff
@@ -0,0 +1,53 @@
...
+            throw new Error(`Invalid setting value for colorVariables: ${ config.style }`);
+        }
+
+        const results = [];
+        ast.first.walk((child) => {
+            // Only evaluate hex colors
+            if (child.type !== 'word' || !/^#/.test(child.value)) {
```
jwilsson: We can use `child.isColor` instead of the regex here.

lib/linters/color_variables.js
```diff
@@ -0,0 +1,53 @@
...
+            results.push({
+                column: node.source.start.column + node.prop.length + node.raws.between.length + (child.source.start.column - 1),
+                message: util.format(this.message, color)
+            });
+        });
+
+        if (results.length) {
+            return results;
+        }
+        return null;
```
jwilsson: Could you remove this? We just let it imply `undefined` if no errors were found.

lib/linters/color_variables.js
```diff
@@ -0,0 +1,53 @@
...
+        const results = [];
+        ast.first.walk((child) => {
+            // Only evaluate hex colors
+            if (child.type !== 'word' || !/^#/.test(child.value)) {
+                return;
+            }
+
+            // If this is a variable assignment, allow hex colors
+            const prop = node.prop;
+            if (/@[a-zA-Z0-9-_]*/.test(prop)) {
```
jwilsson: Could you change this to the new [`isVariable`](https://github.com/lesshint/lesshint/blob/master/docs/developer-guide/api/utils.md#isvariable) util (I just merged it to master).

test/specs/linters/color_variables.js
```diff
@@ -0,0 +1,68 @@
...
+            }];
+
+            return spec.parse(source, function (ast) {
+                const result = spec.linter.lint({}, ast.root.first);
+
+                expect(result).to.deep.equal(expected);
+            });
+        });
+
+        it('should throw on invalid "style" value', function () {
```
jwilsson: This test can be removed since there's no `style` option.

lib/linters/color_variables.js
```diff
@@ -0,0 +1,53 @@
...
+    message: '%s should be replaced with an existing variable.',
+
+    lint: function colorVariablesLinter (config, node) {
+        /**
+         * Just in case, since postcss will sometimes include the semicolon
+         * in declaration values
+         */
+        node.value = node.value.replace(/;$/, '');
+
+        const ast = parser(node.value, {
```
jwilsson: Could you change this to the new [`parseValue`](https://github.com/lesshint/lesshint/blob/master/docs/developer-guide/api/utils.md#parsevalue) util (I just merged it to master).

WattyRev:
@jwilsson, no worries about the delay. I've made the changes you requested.

jwilsson:
Thanks for contributing! 🎉 

I'll publish a new version later this week.

PR #375, opened 2017-04-13
If custom linter is not found by absolute url or relative to cwd, check relative to .lesshintrc

WattyRev:
#372 

Basic gist is that I am adding _path to the config data that contains the path to the .lesshintrc file. This allows me to find custom linters relative to .lesshintrc.

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -15,7 +15,7 @@ const loadConfig = function (path) {
     }
 
     data = JSON.parse(data);
-    data._path = path;
+    data.configPath = path;
     return data;
 };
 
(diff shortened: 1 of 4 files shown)


WattyRev:
*bump*

lib/config-loader.js
```diff
@@ -14,7 +14,9 @@ const loadConfig = function (path) {
         data = data.slice(1);
     }
 
-    return JSON.parse(data);
+    data = JSON.parse(data);
+    data._path = path;
```
shellscape: I'm personally not a fan of properties led with an underscore. I think `data.filePath` or `data.configPath` would be good and semantic here

WattyRev: Updated.

shellscape: cool. gotta let the others chime in before I can merge :)

jwilsson:
@WattyRev Thanks for contributing! 🎉 

I'll see if I can get a new release out over the weekend.

PR #378, opened 2017-04-22
Add Lesshint.checkFiles() method

jwilsson:
This is a more general method which works on globs and handles everything instead of having separate methods.

Point 4 of #357.

I'm thinking we should deprecate these methods in favor of `Lesshint.checkFiles()`:
* `Lesshint.checkDirectory()`
* `Lesshint.checkFile()`
* `Lesshint.checkPath()`

I also think we should deprecate the `fileExtensions` and `excludedFiles` options and let them be handled by glob patterns instead. What do other people think?

Is it enough to add docs about things being deprecated or should we use something like [`util.deprecate`](https://nodejs.org/api/util.html#util_util_deprecate_function_string)?

### DIFF
--- a/docs/developer-guide/api/lesshint.md
+++ b/docs/developer-guide/api/lesshint.md
@@ -38,6 +38,19 @@ result.then((results) => {
 });
 ```
 
+## `Lesshint.checkFiles(pattern)`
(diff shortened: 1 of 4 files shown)


jwilsson:
Closing in favor of #440.

PR #454, opened 2018-02-07
fix(doc): change paths to args

mbrechet:
``` args``` is the right parameter to set less files to runner


jwilsson:
Nice catch, thanks!

PR #467, opened 2018-04-23
zeroUnit: exclude flex-grow/-shrink

benmosher:
these rules deal in unitless numbers; to suggest they require units is an error.
see: https://developer.mozilla.org/en-US/docs/Web/CSS/flex

would ignore `flex` too, but it has `flex-basis` as its third input, which may need a unit.

none


jwilsson:
Thanks for contributing! 🎉  I'll get a new release out right away.

Issue #495, opened 2018-10-23
Property order concentric


Tiross:
Hello, 

It will be nice to choose an other property order like [Concentric CSS](https://rhodesmill.org/brandon/2011/concentric-css/).

PR #496, opened 2018-10-23
Concentric

Tiross:
#495


.editorconfig
```diff
@@ -0,0 +1,16 @@
+
```
jwilsson: Not really sure why this file was added.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+
+## Unreleased
```
jwilsson: Thanks, but I'll update the changelog on release.

lib/linters/property_ordering.js
```diff
@@ -8,12 +8,20 @@ module.exports = {
...
-        if (config.style !== 'alpha') {
+        let ordering;
+
+        try {
+            ordering = require('../config/ordering/' + config.style);
+        } catch (excep) {
             throw new Error(`Invalid setting value for propertyOrdering: ${ config.style }`);
         }
 
+        if (config.style !== 'alpha') {
```
jwilsson: I think we can remove this check and use `util.format` when reporting the error instead. Take a look at `stringQuotes` for example.

.editorconfig
Tiross: My bad, it's not related to the topic, but I think it's not a bad idea, it change IDE behaviour to match what is in the file. 

Example, I prefer indent with 2 spaces, with that my Atom automatically change for 4 spaces for this project.

CHANGELOG.md
Tiross: Removed

lib/linters/property_ordering.js
Tiross: Yep, seems better

.editorconfig
jwilsson: That's true. Let's leave it in then!

jwilsson:
Thanks for contributing! 🎉 

I'll get a new version out straight away.

PR #497, opened 2018-10-24
Better report on ordering errors

Tiross:
Following  #496 

Sorry I did not see it before but it's now really boring to only have a "not concentric" error. 

Here I show a "move it before" message, it's not perfect as you got it for every line, but it's a start. 

I will look later if we can find the correct position on first time.

(It may be better to merge this in a 5.2.1 and create a feature request for later)


jwilsson:
No worries! Thanks for the quick fix! 😀 

Could you open an issue for tracking the feature request?

jwilsson:
And released in `5.2.1`.

Issue #480, opened 2018-05-28
[new rule] check colon in pseudo-classes and pseudo-elements


wareczek:
[Pseudo Classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) should always have `1` colon.

valid:
```less
&:active {
```
invalid:
```less
&::active {
```

---

[Pseudo Elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) should always have `2` colons.

valid:
```less
&::after {
```
invalid:
```less
&:after {
```

PR #499, opened 2018-10-28
New rule "pseudo-elements"

Tiross:

#480 

* I added a watch task to simplify development, it run a test task on every js file modification. 
* I only wanted to try to add a new rule, I never used postcss parser before, so my code may not be so optimal
* Pseudo elements list is from MDN

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,4 +28,8 @@ gulp.task('coveralls', ['test'], () => {
         .pipe(coveralls());
 });
 
+gulp.task('watch', () => {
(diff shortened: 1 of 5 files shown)


jwilsson:
Thanks for contributing! 🎉 

Let's wait a few days and see if AppVeyor adds Node 11 (that's why it's failing there). Otherwise, we can merge it anyway and test it manually on Windows.

jwilsson:
Thanks again! 🎉 

I've just published `5.3.0` including this rule.

